PR #179, opened 2014-05-13
Custom reporters

nathan:
**Note**: Shane asked me to make this PR so we could discuss custom reporters. As I said earlier:

> I think it would be a cool feature to have, but the way I implemented it definitely wasn't the best.
> And we should think about the design a lot more carefully if we actually want it in Scratch.

That said, this is quite fun to play around with.

---

Rewrote the interpreter and added the user interface to support custom reporters.

screen shot 2014-05-11 at 14 36 56

screen shot 2014-05-11 at 12 58 21

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -379,7 +379,7 @@ public class Block extends Sprite {
 		for (i = 0; i < labelsAndArgs.length; i++) {
 			item = labelsAndArgs[i];
 			// Next line moves the argument of if and if-else blocks right slightly:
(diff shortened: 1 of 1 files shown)


liam4:
*closes without reason* @sclements :P

BookOwl:
PLEASE ADD THIS @thisandagain

DrKat123:
Yeah

DrKat123:
Yeah

NoMod-Programming:
PLZ ADD THIS!!!

thisandagain:
We will consider adding this in Scratch 3.0

PR #183, opened 2014-05-14
Fixed shadow disappearing when a script runs while being dragged

### LINKED ISSUE #132: Shadow is removed when a hat runs while being dragged
<a href="https://github.com/nathan"></img></a> **Issue by [nathan](https://github.com/nathan)**
_Saturday Apr 05, 2014 at 22:29 GMT_
_Originally opened as https://github.com/LLK/scratch-flash-common/issues/1405_

---

If you press the space key while dragging a when space key pressed block the shadow is removed from the block.

nathan:
Fixes #132.

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -298,7 +298,7 @@ public class Block extends Sprite {
 	}
 
 	public function showRunFeedback():void {
-		if (filters.length > 0) {
+		if (filters && filters.length > 0) {
 			for each (var f:* in filters) {
 				if (f is GlowFilter) return;
 			}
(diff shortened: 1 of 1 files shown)


src/blocks/Block.as
```diff
@@ -298,13 +298,22 @@ public class Block extends Sprite {
 	}
 
 	public function showRunFeedback():void {
-		if (!filters || (filters.length == 0)) {
-			filters = runFeedbackFilters();
+		if (filters.length > 0) {
```
sclements: It seems that this line needs to make sure that 'filters' is not null.

nathan: Fixed.

Issue #213, opened 2014-05-28
Optimisations when costume, rotation, size, effects don't change


TheLogFather:
Sorry this is rather long - but I'm looking for a bit of guidance...

One of the most common ways Scratchers write script is using loops which check certain conditions, and always perform some looks operation(s) depending on the condition. (E.g. "forever / if something then / switch to costume 1 / else / switch to costume 2 ...") This is currently very expensive, since Scratch always performs all the calculations necessary for the costume switch, even if it's already showing that costume! I think optimising this away would be a big help to Scratchers.

I've added checks which ensure such calculations are not performed if the costume, or whatever, isn't actually changing. As an example, this reduces the time for a few thousand costume 'non-switches' from over a second down to <100th sec.
One complication I found, which makes the change a little more messy than I'd hoped, was that when a sprite is first initialised, its values are set first, and then the costumes are read from file/server and it calls the appropriate costume/looks routines to get everything showing appropriately.

However, that means when it first calls those looks routines, the values are already set correctly, so simply adding checks whether some value changes means it doesn't set up the bitmap info correctly first time through.

I added a "doneInit" variable for a sprite, which starts off false, and is set to true eventually (once it has finished instantiating stuff n whatever way). It can then only do the above-mentioned optimisation if doneInit is true.

The problem is that I found there's more than one route it can take to get to the 'finished setup' point, hence more than one place where doneInit should be set true. I can therefore imagine there may be other rare cases I've missed, or there may even be more places added at some later point.
So, I'm wondering if anyone has any better suggestions how to go about this, or some suggestion how things could be rearranged to avoid this complication...? 

Thanks for reading this far!

PR #230, opened 2014-05-30
Avoid unnecessary work for non-changes of costume, size, direction. And fix rotated sprite touching issues.

TheLogFather:
See #213 for some info/background on the 'non-changes' optimisation.
See also: http://scratch.mit.edu/discuss/post/23532/

Here's a few refs for the touching check problems:
http://scratch.mit.edu/discuss/post/282216/
http://scratch.mit.edu/discuss/topic/30769/
http://scratch.mit.edu/discuss/topic/31853/
http://scratch.mit.edu/discuss/topic/31153/


src/scratch/ScratchSprite.as
```diff
@@ -222,7 +233,9 @@ public class ScratchSprite extends ScratchObj {
 		var wasFlipped:Boolean = isCostumeFlipped();
 		d = d % 360;
 		if (d < 0) d += 360;
-		direction = (d > 180) ? d - 360 : d;
+		d = (d > 180) ? d - 360 : d;
+		if (skipNoChange && d == direction) return; // TLF: nothing changed
+        direction = d;
```
TheLogFather, line 238: Drat those tabs!

sclements:
Hi @TheLogFather, @cwillisf and I took a look at this and we're concerned about adding additional state variables. Do you think this optimization can be achieved without the new state variable?

TheLogFather:
I'd quite like to know exactly why it needed it...
I wrote quite a bit about it in https://github.com/LLK/scratch-flash/issues/213, but it was so long ago now that I definitely don't remember what was going on (and I didn't really get to the bottom of it anyway, IIRC).
It seemed like the sprite's state was somehow getting set to the correct values without the costume really being 'set up' - so when it came to do a switch again, i.e. when it *really* sets it up, it wasn't making the change because it appeared to already have the right state.
If that could be fixed then the extra variable shouldn't be needed...

PR #251, opened 2014-06-09
Fix hiding the 'say' bubble WRT number formatting

cwillisf:
When setting bubble text, numbers are rounded to two decimal places if
necessary. The check before hiding the bubble that makes sure the
current bubble text matches the current value -- probably intended to
make sure it's "our" bubble -- now uses the same text formatting code to
avoid false negatives.

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -818,10 +818,10 @@ public class Scratch extends Sprite {
 
 	public function handleTool(tool:String, evt:MouseEvent):void { }
 
-	public function showBubble(text:String, x:* = null, y:* = null, width:Number = 0):void {
+	public function showBubble(text:String, source:Object, x:* = null, y:* = null, width:Number = 0):void {
 		if (x == null) x = stage.mouseX;
 		if (y == null) y = stage.mouseY;
-		gh.showBubble(text, Number(x), Number(y), width);
+		gh.showBubble(text, source, Number(x), Number(y), width);
 	}
 
 	// -----------------------------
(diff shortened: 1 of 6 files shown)


src/Scratch.as
```diff
@@ -818,10 +818,10 @@ public class Scratch extends Sprite {
 
 	public function handleTool(tool:String, evt:MouseEvent):void { }
 
-	public function showBubble(text:String, x:* = null, y:* = null, width:Number = 0):void {
+	public function showBubble(text:String, source:Object, x:* = null, y:* = null, width:Number = 0):void {
```
nathan: `source` really doesn't make sense in this `showBubble` since you always create a new `TalkBubble` and never check the source of the old one. Maybe you should just pass `null`?

cwillisf: Hmm... I could see it going either way. I set the source so that you can always trust that it would be set to a meaningful value; it's possible that we might want to check it in the future for some reason, or the value might be helpful in debugging.
Can you think of a reason it would be harmful or misleading to set the source like this? My understanding of the current code is that the value set in this case won't be used, but also won't hurt anything.

nathan: "We might use it later" is not a good reason to add something now. If we ever need to set the source for app bubbles, we can make that change when we need it.

cwillisf: In my mind the balance is this:
- Arguments for: consistency (major), might be useful (minor)
- Arguments against: currently not necessary (medium)

If I were adding the 'source' variable just for this case then I would totally agree that it's a bad idea to make a change just because it might be useful some day. However, since some cases already need this variable, it makes me uncomfortable to intentionally set it to a "wrong" value (null) just because we don't currently need it to be correct. I feel like it's a least-astonishment violation.

PR #260, opened 2014-06-16
Prevent creating a global with the same name as a local

### LINKED ISSUE #119: Should not be able to create global and local variables of the same name
<a href="https://github.com/nathan"></img></a> **Issue by [nathan](https://github.com/nathan)**
_Tuesday Feb 25, 2014 at 22:33 GMT_
_Originally opened as https://github.com/LLK/scratch-flash-common/issues/1367_

---

You can currently do this by making a sprite-local variable, then switching to the stage and making a global variable.

### LINKED ISSUE #349: Don't allow variables and lists with the same name
Scratch 1.4 did not allow a variable to have the same name as a list. Scratch 2.0 should probably not allow this either, since it makes the reporter blocks confusingly similar.

screen shot 2014-07-08 at 16 41 36

### LINKED ISSUE #374: Renaming variable can shadow other variables
If you rename a global to the same name as a local in a different sprite, that sprite can have two variables of the same name.

nathan:
This also adds a dialog box explaining why the variable could not be created and prevents creating a list with the same name as a variable.

Fixes #119.
Fixes #349.
Fixes #374.

### DIFF
--- a/src/scratch/PaletteBuilder.as
+++ b/src/scratch/PaletteBuilder.as
@@ -174,7 +174,7 @@ public class PaletteBuilder {
 
 	protected function createVar(name:String, varSettings:VariableSettings):* {
 		var obj:ScratchObj = (varSettings.isLocal) ? app.viewedObj() : app.stageObj();
-		if (obj.anyChildOwnsVar(name)) {
+		if (obj.hasVarName(name)) {
 			DialogBox.notify("Cannot Add", "That variable name is already in use.");
 			return;
 		}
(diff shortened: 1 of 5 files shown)


src/scratch/PaletteBuilder.as
```diff
@@ -174,6 +174,10 @@ public class PaletteBuilder {
 
 	protected function createVar(name:String, varSettings:VariableSettings):* {
 		var obj:ScratchObj = (varSettings.isLocal) ? app.viewedObj() : app.stageObj();
+		if (obj.anyChildOwnsVar(name)) {
+			DialogBox.notify("Cannot Add", "That variable name is already in use.");
+			return;
+		}
 		var variable:* = (varSettings.isList ? obj.lookupOrCreateList(name) : obj.lookupOrCreateVar(name));
```
cwillisf, line 181: If you change this to use lookup early and create later instead of using lookupOrCreate, you could provide the same feedback in the case where you try to create a sprite variable with the same name as an existing global variable. Currently that case offers no feedback: it just silently refuses to create a new variable due to lookupOrCreate.

nathan, line 181: Fixed.

PR #336, opened 2014-07-08
Allow only one HTTP poll at a time per extension

cwillisf:
Without this change, long-running poll requests (such as those resulting in a timeout) can cause arbitrarily long request queues, eventually leading to spontaneous termination of the application.

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -29,6 +29,7 @@ import blocks.BlockArg;
 import flash.events.*;
 	import flash.external.ExternalInterface;
 	import flash.net.*;
+	import flash.utils.Dictionary;
 	import flash.utils.getTimer;
 	import blocks.Block;
 	import interpreter.*;
@@ -43,6 +44,7 @@ public class ExtensionManager {
 	private var app:Scratch;
 	private var extensionDict:Object = new Object(); // extension name -> extension record
 	private var justStartedWait:Boolean;
+	private var pollInProgress:Dictionary = new Dictionary(true);
 	static public const wedoExt:String = 'LEGO WeDo';
 
 	public function ExtensionManager(app:Scratch) {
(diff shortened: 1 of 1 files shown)


src/extensions/ExtensionManager.as
```diff
@@ -497,16 +499,27 @@ public class ExtensionManager {
 	}
 
 	private function httpPoll(ext:ScratchExtension):void {
+
+		if (pollInProgress[ext]) {
```
cwillisf, line 503: The dictionary key could be the URL instead of the extension object, but this allows a slightly earlier out and seems like it shouldn't make a difference.

src/extensions/ExtensionManager.as
```diff
@@ -497,16 +499,27 @@ public class ExtensionManager {
...
+
 		// Poll via HTTP.
 		function completeHandler(e:Event):void {
+			delete pollInProgress[ext];
 			processPollResponse(ext, loader.data);
 		}
-		function errorHandler(e:Event):void { } // ignore errors
+		function errorHandler(e:Event):void {
+			delete pollInProgress[ext];
+		} // ignore errors
```
nathan: It might make sense to put the comment inside of the function now that it has a body.

cwillisf: Heh, yeah, might as well. :)

src/extensions/ExtensionManager.as
cwillisf, line 503: To clarify "seems like it shouldn't make a difference": the only situation in which it would make a difference is if two different extensions were polling the same URL, which shouldn't ever happen. Even if it does happen, we'd just have up to a handful of polls queued... still less than the tens of thousands that I was sometimes seeing without this change.

src/extensions/ExtensionManager.as
```diff
@@ -497,16 +499,28 @@ public class ExtensionManager {
 	}
 
 	private function httpPoll(ext:ScratchExtension):void {
+
```
nathan, line 502: Is that blank line intentional?

PR #348, opened 2014-07-08
Translation fixes

nathan:
Fixes some minor translation issues.

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -771,7 +771,7 @@ public class Scratch extends Sprite {
 			return;
 		}
 		var d:DialogBox = new DialogBox();
-		d.addTitle(Translator.map('Save project') + '?');
+		d.addTitle('Save project?');
 		d.addButton('Save', save);
 		d.addButton('Don\'t save', proceedWithoutSaving);
 		d.addButton('Cancel', cancel);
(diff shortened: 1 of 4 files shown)


src/Scratch.as
```diff
@@ -771,7 +771,7 @@ public class Scratch extends Sprite {
 			return;
 		}
 		var d:DialogBox = new DialogBox();
-		d.addTitle(Translator.map('Save project') + '?');
+		d.addTitle('Save project?');
```
cwillisf, line 775: This will change the translator token from "Save project" to "Save project?"... I actually think that's a good change, but it seems like it will require corresponding changes to the translation files.

nathan, line 775: Yeah, it will, but it was an error to translate "Save project" and then append a question mark.

cwillisf, line 775: I agree (that's what I meant by "I actually think that's a good change"), I'm just concerned about making sure the translation files change too. I don't know much about our translation system yet... are the translation files in this repository? Could those changes be bundled with this one to make it atomic?

nathan, line 775: Unfortunately, no. The translation files are in [LLK/scratchr2_translations](https://github.com/LLK/scratchr2_translations/), which is periodically updated from [translate.scratch.mit.edu](http:/translate.scratch.mit.edu/). But we don't have much to worry about, since the string "Save project" [isn't in the editor.po anyway](http://translate.scratch.mit.edu/pt/editor/editor.po/translate/?search=Save%20project&sfields=source&sfields=target). I've added the two missing strings ("Don't save" and "Save project?") to the app's `strings()` method, but that's all I can do inside this PR.

cwillisf, line 775: Gotcha. Thanks for the explanation!

PR #416, opened 2014-07-29
Fix for parameter focusing

### LINKED ISSUE #414: Focus new parameters in procedure editor
When a label is added in the procedure editor, it correctly receives keyboard focus so the user can begin typing the label. When a parameter is added, it should be focused too.

natecraddock:
Fixes #414

Try it here: https://dl.dropboxusercontent.com/u/76236043/scratch/parameterFocus.swf

I also changed the code so when you re-open a procedure to edit it, the first parameter receives keyboard focus, not the first text object as the code was previously doing.

(My first bugfix, yay!)

### DIFF
--- a/src/ui/ProcedureSpecEditor.as
+++ b/src/ui/ProcedureSpecEditor.as
@@ -292,10 +292,7 @@ public class ProcedureSpecEditor extends Sprite {
 		addChild(o);
 		if (stage) {
 			if (o is TextField) stage.focus = TextField(o);
-			if (o is BlockArg) {
-				var arg:BlockArg = o as BlockArg;
-				arg.startEditing();
-			}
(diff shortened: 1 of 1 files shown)


src/ui/ProcedureSpecEditor.as
```diff
@@ -292,7 +292,10 @@ public class ProcedureSpecEditor extends Sprite {
 		addChild(o);
 		if (stage) {
 			if (o is TextField) stage.focus = TextField(o);
-			if (o is BlockArg) stage.focus = BlockArg(o).field;
+			if (o is BlockArg) {
+				var arg:BlockArg = o as BlockArg;
+				arg.startEditing();
```
nathan: Might be cleaner just to do `BlockArg(o).startEditing();`.

PR #464, opened 2014-08-19
Update Specs.as

DevanWolf:
Updating Block Specs.as to bring back certain blocks and update/fix a few blocks.


sclements:
We're very careful about any additions to the language and every change undergoes a lot of consideration. This change cannot go in at this time.

PR #488, opened 2014-09-12
Include JSON in error report during JSON parse

cwillisf:
This could get a little verbose but will hopefully help nail down exactly where JSON parsing errors are coming from


src/util/JSON.as
```diff
@@ -49,12 +49,13 @@ public class JSON {
 	public static function parse(s:String):Object {
 		// Parse the JSON string and return the resulting object.
 		var json:util.JSON = new util.JSON();
+		json.buf = s;
```
cwillisf, line 52: 'buf' was only being used in the stringify() call tree before

Issue #206, opened 2014-05-23
Upload of MP3 files results in blank clip


Paddle2See:
Many times, when an MP3 sound file is uploaded to a sprite, the conversion process will appear to go okay, but the result is a blank sound clip.  No errors are produced.

I'm pretty sure I reported this bug a long time ago - but I can't find it so I'm reporting it again.  Sorry for the duplicate, if it is one :/

I have a sample MP3 file that produces this result, if you need one.

PR #558, opened 2014-12-03
Use Flash 11's Sound::loadCompressedDataFromByteArray to load MP3 files

sclements:
When available, use this method to load MP3s instead of our custom build-a-swf-in-memory method. For those with newer Flash versions this will fix https://github.com/LLK/scratch-flash/issues/206.

### DIFF
--- a/src/ui/media/MediaLibrary.as
+++ b/src/ui/media/MediaLibrary.as
@@ -673,11 +673,16 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
 			SCRATCH::allow3d {
 				sound = new Sound();
 				data.position = 0;
-				sound.loadCompressedDataFromByteArray(data, data.length);
-				MP3Loader.extractSamples(name, sound, sound.length * 44.1, function (out:ScratchSound):void {
-					snd = out;
-					startSoundUpload(out, origName, uploadComplete);
-				});
+				try {
+					sound.loadCompressedDataFromByteArray(data, data.length);
+					MP3Loader.extractSamples(name, sound, sound.length * 44.1, function (out:ScratchSound):void {
+						snd = out;
+						startSoundUpload(out, origName, uploadComplete);
+					});
+				}
+				catch(e:Error) {
+					uploadComplete();
+				}
 			}
 
 			if (!sound)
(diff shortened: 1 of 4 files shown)


src/ui/media/MediaLibrary.as
```diff
@@ -669,12 +669,24 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
...
-			setTimeout(function():void {
-				MP3Loader.convertToScratchSound(sndName, data, function(s:ScratchSound):void {
-					snd = s;
-					startSoundUpload(s, origName, uploadComplete);
+			var sound:Sound;
+			SCRATCH::allow3d {
+				sound = new Sound();
+				data.position = 0;
+				sound.loadCompressedDataFromByteArray(data, data.length);
+				MP3Loader.extractSamples(name, sound, sound.length * 44.1, function (out:ScratchSound):void {
```
cwillisf: Is it always safe to use 44.1 here, even (for example) with a 22050Hz MP3?

src/ui/media/MediaLibrary.as
```diff
@@ -669,12 +669,24 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
...
 			if (app.lp) app.lp.setTitle('Converting mp3...');
-			setTimeout(function():void {
-				MP3Loader.convertToScratchSound(sndName, data, function(s:ScratchSound):void {
-					snd = s;
-					startSoundUpload(s, origName, uploadComplete);
+			var sound:Sound;
+			SCRATCH::allow3d {
+				sound = new Sound();
+				data.position = 0;
+				sound.loadCompressedDataFromByteArray(data, data.length);
```
cwillisf: From the docs I see that loadCompressedDataFromByteArray can throw. Does it need to be wrapped in a try/catch or would it be safe to throw here?

src/ui/media/MediaLibrary.as
sclements: Yes, if you read the docs Sound objects always extract audio at 44100hz

src/ui/media/MediaLibrary.as
sclements: Good call. Better now?

PR #574, opened 2014-12-18
Scale the eraser up to make it easier to use

cwillisf:
The bitmap-mode eraser tool is now 4x larger than it used to be.


sclements:
:+1:

PR #577, opened 2014-12-24
Check for HTTPS also when determining if offline

djdolphin:
When viewing a project on https://scratch.mit.edu/, the project editor is in the space where the player should be, and see inside is broken:
screen shot 2014-12-24 at 8 56 25 am
This has been reported multiple times on the forums, and checking for both HTTP and HTTPS when determining whether to default to the player or not should fix the problem.


sclements:
:+1:  (could be shorter logic but this works)

PR #592, opened 2015-01-22
Force an update to the 3d renderer when a sprite is added to the stage.

sclements:
This fixes an issue where recently added bitmap sprites appear invisible when the 3d renderer is in use.


cwillisf:
:+1:

PR #597, opened 2015-01-27
Store eraser width separately from pen width

### LINKED ISSUE #596: Minimum Eraser size is now bigger
Server and offline: introduced in scratch.mit.edu v431

any browser
any os

In the Editor, the eraser is now 4x bigger. The minimum size of the eraser used to be 1px wide and now it is 4px wide

http://scratch.mit.edu/discuss/topic/83654/

cwillisf:
The default eraser size is larger than the default pen size, but the limits are currently the same.
This fixes #596


sclements:
:+1:

PR #605, opened 2015-02-01
Comment layout fix

AndersBillLinden:
When a user defined block has been edited, and some inputs has been added/removed, the width of the block can change, so we need to call fixCommentLayout.


sclements:
Thanks!

PR #606, opened 2015-02-01
Save needed fix

AndersBillLinden:
If the user saves the project and then starts a new, he/she is prompted to save the project.


sclements:
This is by design. We want to give the user a chance to potentially remix the current project that they have altered. Alternatively, if it's already their project, on the website, we'd like them to know that it hasn't saved there. This flag is used primarily for 'online' editing.

PR #607, opened 2015-02-01
Automatically select text when adding comment

AndersBillLinden:
When the user adds a comment to a block, it would be convenient to have the text "add comment here..." selected and the TextField focused.


sclements:
I agree!

PR #610, opened 2015-02-01
Allowing sprites outisde stage

AndersBillLinden:
Sprites should be able to go outside the stage. Else, a lot of effects are impossible. In my remix of Dog-a-thon at http://scratch.mit.edu/projects/45942694/ , the stones should not all start at the left edge of the screen, but at an x-coordinate to the left of the edge. They should also be able to leave the right edge. With the current version that is impossible to do because of the function ScratchSprite.keepOnStage.

### DIFF
--- a/src/primitives/MotionAndPenPrims.as
+++ b/src/primitives/MotionAndPenPrims.as
@@ -293,7 +293,6 @@ public class MotionAndPenPrims {
 		var oldX:Number = s.scratchX;
 		var oldY:Number = s.scratchY;
 		s.setScratchXY(newX, newY);
-		s.keepOnStage();
 		if (s.penIsDown) stroke(s, oldX, oldY, s.scratchX, s.scratchY);
 		if ((s.penIsDown) || (s.visible)) interp.redraw();
 	}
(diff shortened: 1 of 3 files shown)


griffpatch:
I was thinking that sprites could be only kept on stage when the program is
not running? That way whenever it stood all the sprites come back?
Alternatively a check box on the sprites details page labeled allow off
screen so a user can title that option on of they want it... I would love
that one! :)

AndersBillLinden:
griffpatch I created a new branch with a correct spelling, please comment there :)

AndersBillLinden:
Yeah, maybe keeping them on stage when the program is not running is a good thing, or at least near the viewport.

ntlrsk:
We can see how this would be very valuable in many projects, however there are many design questions to consider - such as are there any limits to the x y coordinates? Is it OK that the same script (forever, move 10 steps) will do different things depending on whether a box in the info window is checked or not? How will "if on edge bounce" respond if the sprite starts off the screen -- will it bounce out of view or come into view? Should we add a block that controls whether they stay onstage or not, so that it can be more programmatically obvious and controlled (as we did for the previously hidden rotation style feature)? Those are some of the questions...

AndersBillLinden:
@ntlrs Please, post in the open branch!

PR #611, opened 2015-02-01
Allowing sprites to go outside or partly outside the stage.

AndersBillLinden:
Sprites should be able to go outside the stage. Else, a lot of effects are impossible. In my remix of Dog-a-thon at http://scratch.mit.edu/projects/45942694/ , the stones should not all start at the left edge of the screen, but at an x-coordinate to the left of the edge. They should also be able to leave the right edge. With the current version that is impossible to do because of the function ScratchSprite.keepOnStage.

### DIFF
--- a/src/primitives/MotionAndPenPrims.as
+++ b/src/primitives/MotionAndPenPrims.as
@@ -293,7 +293,6 @@ public class MotionAndPenPrims {
 		var oldX:Number = s.scratchX;
 		var oldY:Number = s.scratchY;
 		s.setScratchXY(newX, newY);
-		s.keepOnStage();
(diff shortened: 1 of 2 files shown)


griffpatch:
I was thinking that sprites could be only kept on stage when the program is
not running? That way whenever it stood all the sprites come back?
Alternatively a check box on the sprites details page labeled allow off
screen so a user can title that option on of they want it... I would love
that one! :)

AndersBillLinden:
@griffpatch I think it is a good idea!

sclements:
This is a dramatic change to Scratch. We've been discussing an alternative which relies on a property of the Stage. So, some stages will have bounds and others will be unbounded. Here is our plan: http://scratch.mit.edu/discuss/topic/52301/?page=3

PR #617, opened 2015-02-07
When key released event block.

AndersBillLinden:
A block for key release events. If you want to create a game where objects are maneuvered with the keyboard, this would be appropriate so that they can stop when keys are released. Translations will also be needed.

### DIFF
--- a/src/Specs.as
+++ b/src/Specs.as
@@ -220,6 +220,7 @@ public class Specs {
 		// triggers
 		["when @greenFlag clicked",				"h", 5, "whenGreenFlag"],
 		["when %m.key key pressed",				"h", 5, "whenKeyPressed", 		"space"],
+		["when %m.key key released",				"h", 5, "whenKeyReleased", 		"space"],
 		["when this sprite clicked",			"h", 5, "whenClicked"],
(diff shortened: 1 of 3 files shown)


sclements:
Interesting idea. As I'm sure you know, the missing release hat can be worked around by using the <key X pressed> reporter.

AndersBillLinden:
Maybe we should also add an option to key press to exclude repetitions, defaulting to not excluding those

AndersBillLinden:
Ok, I have to admit that I do not know about the reporter! What is it?

AndersBillLinden:
@sclements Ok, I have to admit I have not heard about the reporter, what is it? Also, would github send you a notice even if I did not prefix this comment with an at-sign and your name?

sclements:
Yes, I received a notice even without being tagged. You can find the reporter in the Sensing block category. A reporter is a block that returns a value and can be used as an input for another block.

AndersBillLinden:
Ah, yeah, now I see that reporter block. Maybe it makes sense to only have the key press block then. Or "an all keys are released" block in addition to the key press block. In either case, would it not be great to handle the key repetition "dilemma"?

PR #618, opened 2015-02-08
Only have one FileFilter containing both suffixes (.sb and .sb2)

AndersBillLinden:
Most people do not actually care about file suffixes so why not just have one file filter with both suffixes?


sclements:
:+1:

PR #624, opened 2015-02-10
Faster clicking of media library items.

AndersBillLinden:
The use will not have to wait for the visual feedback when clicking an image in the media library.


sclements:
We believe that the performance issue is elsewhere and we'd like to stick to the design (on click).

PR #628, opened 2015-02-11
Move scroll gesture pickiness to the scripts area

cwillisf:
Most ScrollFrames are OK with catching the drag event from a child, so
move the code that prevents that into the scripts part. Any other scroll
frame that wants to prevent it can follow this pattern.


sclements:
:+1:

PR #634, opened 2015-02-17
Media library lazy load

cwillisf:


sclements:
:+1:

2jour:
Does this need qa verification for 2.0.94 @cwillisf ?

cwillisf:
@2jour this is only in the mobile branch for now, not in master.

PR #635, opened 2015-02-18
Fix bugs noticed in today's tablet meeting

cwillisf:
Picking "OK" with no costume selected no longer adds all costumes.
The sound library now shows all sounds instead of just one at a time.


sclements:
we already have these changes now, don't we?

cwillisf:
Yep -- these are the "missing" changes that we reimplemented again this week.

sclements:
doh!

PR #639, opened 2015-02-23
When renaming a variable, the current name is proposed.

AndersBillLinden:
(It would also be nice if the text was selected, but it seems like it can be when deployed, judging from the code comment in DialogBox.showOnStage).


sclements:
:+1:

PR #641, opened 2015-02-24
Use drawWithQuality, don't change stage quality

cwillisf:
Changing the stage quality sometimes triggers a resize event. If this happens while converting a costume from vector to bitmap the image editor is shut down in the middle of the conversion, causing a crash.

I was unable to determine the reason that changing the stage quality sometimes fires a resize event and sometimes doesn't, but using drawWithQuality is probably more efficient anyway.


cwillisf:
Hmm... I'll rework this for compatibility with Flash 10.2...

PR #642, opened 2015-02-24
Convert to bitmap crash fix

cwillisf:
This change fixes a crash that can sometimes happen while converting a vector costume into a bitmap costume. The fix is different depending on whether Scratch is running a newer or older version of Flash.

The problem stems from changing the stage quality when rendering the SVG into a bitmap. Changing the stage quality sometimes -- but not always -- fires a 'resize' event that is caught in Scratch.as, causing the image editor to be shut down in the middle of the conversion. The actual crash happens inside BitmapEdit's saveToCostume() method, where targetCostume is unexpectedly null.

On newer versions of Flash we now use drawWithQuality() to avoid the need to change the stage's quality.

On older versions of Flash we now set a flag to suppress the resize event, and clear that flag after the rendering step completes.

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -71,7 +71,7 @@ public class Scratch extends Sprite {
 	public var render3D:IRenderIn3D;
 	public var isArmCPU:Boolean;
 	public var jsEnabled:Boolean = false; // true when the SWF can talk to the webpage
-	public var ignoreResize:int = 0; // If non-zero, temporarily ignore resize events.
+	public var ignoreResize:Boolean = false; // If true, temporarily ignore resize events.
 
 	// Runtime
 	public var runtime:ScratchRuntime;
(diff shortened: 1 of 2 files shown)


sclements:
:+1:

2jour:
can you add steps to repro @cwillisf ?

cwillisf:
I've had more luck reproducing this in the offline editor using 64-bit Windows 8.1, but theoretically it should affect the online editor and other platforms as well. Here are the steps I've found to be most reliable:
1. Start the offline editor.
2. Maximize the window.
3. Click the "Costumes" tab. This should go to "costume1" of the Scratch Cat sprite.
4. Click the "Convert to bitmap" button.

In my experience that sequence was leading to a crash more than 75% of the time, though I was having trouble getting it to happen that consistently outside of a debugger.

2jour:
Thank you :)

PR #643, opened 2015-02-24
The distance to a non-existent sprite is now 10000 (was 0).

### LINKED ISSUE #343: Distance to (non-existent object) returns 0
In Scratch 1.4, DISTANCE TO returns 10000 if the object doesn't exist. Scratch 2.0 returns 0.

sclements:
Fixes #343


cwillisf:
:+1:

PR #644, opened 2015-02-24
Debug blocks

AndersBillLinden:
Adding some blocks that can be used when the player runs in debug mode.
Those are: trace, sensors/reporters for version, os, player type, language, manufacturer and has mp3.
Maybe this is totally uninteresting so I will not blame you if you close this pull request without excuses.
I learned something about blocks and reporters when writing it however.

### DIFF
--- a/src/Specs.as
+++ b/src/Specs.as
@@ -28,6 +28,7 @@
 package {
 	import flash.display.Bitmap;
 	import assets.Resources;
+	import util.DebugUtils;
 
 public class Specs {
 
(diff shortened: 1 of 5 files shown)


sclements:
This is an interesting concept. However, I'm not sure we want to compile this into our production builds. We'll discuss and get back to you. Thanks!

sclements:
Looks like we won't be including this. Perhaps some of these blocks can be implemented as an experimental extension though.

PR #645, opened 2015-02-24
Allow media library scaling for physical size

cwillisf:


sclements:
:+1:  I almost missed that static initialization.

PR #652, opened 2015-02-26
Media library refactor

sclements:
Use separate files for sprites, costumes, backdrops, and sounds. This makes them load faster and the costume library is now generated from the sprite library.


cwillisf:
:+1:

PR #640, opened 2015-02-23
Renaming of lists

AndersBillLinden:
This is a proposal of how to rename lists.

http://scratch.mit.edu/discuss/topic/271/
http://scratch.mit.edu/discuss/topic/301/
http://scratch.mit.edu/discuss/topic/42370/
http://scratch.mit.edu/discuss/topic/46622/
http://scratch.mit.edu/discuss/topic/52360/
http://scratch.mit.edu/discuss/topic/65530/
https://github.com/LLK/scratch-flash/issues/375

PR #654, opened 2015-03-01
Renaming broadcasts (using right click menu)

AndersBillLinden:
Renaming broadcasts not allowing already existing broadcast messages.

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -213,14 +213,6 @@ public class Block extends Sprite {
 		fixArgLayout();
 	}
 
-	public function get isBroadcastSender():Boolean {
-		return op == 'broadcast:' || op == 'doBroadcastAndWait';
-	}
-
-	public function get isBroadcastReceiver():Boolean {
-		return op == 'whenIReceive';
-	}
-
 	public function get broadcastMsg():String {
 		for each (var arg:BlockArg in args) {
 			if (arg.menuName == "broadcast") {
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -213,6 +213,32 @@ public class Block extends Sprite {
 		fixArgLayout();
 	}
 
+	public function get isBroadcastSender():Boolean {
+		return op == 'broadcast:' || op == 'doBroadcastAndWait';
+	}
+
+	public function get isBroadcastReceiver():Boolean {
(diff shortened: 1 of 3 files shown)


sclements:
Thanks for this effort @AndersBillLinden! Unfortunately getters are expensive in AS3 (http://stackoverflow.com/questions/11357427/better-working-with-setter-and-getter/11357513#11357513) and we should avoid using them in the interpreter. If you could remove them from this change I would be happy to add this feature.

AndersBillLinden:
I stepped backwards one step now.

AndersBillLinden:
I also removed getter+setter from the rename-lists branch which pull request is on https://github.com/LLK/scratch-flash/pull/640

AndersBillLinden:
Would you please do the changes that are needed for satisfaction? You can add them to my changes or you can create them yourself. I dislike micro optimizations and prefer code readability. My motto is: If a performance drawback cannot be experienced by the end user, it is not a performance drawback. Still I want to contribute to the project. If all I would contribute with is to initiatiate an implementation of a certain functionality, I am satisfied contributing with that. The possibility of renaming lists is an example. That is something that the Scratch Community has asked for quite a while. I could have written a forum entry like everyone else to try push the change. Instead I did prove that it is not only possible but easily implemented, even as a project outsider. Use what you want with my changes. You can throw them away. You can add a zillion of micro optimzations in them. The only thing that is important to me is something happens.

AndersBillLinden:
Are you sure the compiler will not optimize small functions and make them inline automatically?

AndersBillLinden:
Is there some swf optimizer to use to get the performance you want?

sclements:
No, we don't use a SWF optimizer. To see the performance bottlenecks we use Adobe Scout (https://creative.adobe.com/products/scout) which gives us detailed information about where the Flash player is spending its time.

AndersBillLinden:
I must try it, maybe I change the branch soon.

AndersBillLinden:
I have changed the branch, but there is another getter/setter pair that I want to know about. Should I rewrite that as well? Seems "messy" to do so.

sclements:
This looks good now. We'll get it into the next release (436).

AndersBillLinden:
:+1:

sclements:
@2jour can you test this? I'll put it on a port.

PR #666, opened 2015-03-11
Fix 3d issues 

sclements:
Fixes two major 3d bugs and a few minor primitive bugs due to missing null checks.


cwillisf:
:+1:
Hopefully one of us will soon find time to do a proper fix rather than a workaround...

PR #670, opened 2015-03-13
No longer creating closures with every call to startEdgeTriggeredHats.

sclements:
This drastically improves performance with many clones.

### DIFF
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -21,25 +21,25 @@
 // John Maloney, September 2010
 
 package scratch {
-	import flash.display.*;
-	import flash.events.*;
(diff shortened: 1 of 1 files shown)


griffpatch:
This sounds exciting...  What is classed as a start edge triggered hat?

cpseager:
@sclements 

     } else if ('whenSensorConnected' == hat.op) {

Is this section dead old 'picoboard' code no longer required at all?

sclements:
@cpseager yes, we could probably remove that.
@griffpatch loudness, timer, and video motion hat blocks.

This method is called on all hat blocks for each sprite and clones.

cwillisf:
:+1:

Issue #369, opened 2014-07-11
Make blocking HTTP reporters use the polling mechanism


nathan:
Currently, blocking HTTP reporters (`R`-type blocks in HTTP extensions) send an HTTP request and wait for its response; when the request finishes, execution continues, using the response as the value for the reporter.

However, it would be more consistent (and help to reduce the number of concurrent requests) if they instead sent a request which terminated quickly and checked its status via the `poll` command. The documentation [suggests](http://wiki.scratch.mit.edu/w/images/ExtensionsDoc.HTTP-9-11.pdf#page=6) the following:

```
_result 7639 82
```

(where `7639` is a request ID and `82` is the result of the operation.) This seems like the opposite of `_busy`, though: `_busy` must be sent with every poll response until the command should terminate, whereas `_result` must be sent with only a single poll response when the reporter should terminate.

la10736:
There is a way to track the story of that feature? I would like to know when is complete or contribute to it if you are too busy to implement it.

la10736:
I would like implement this feature. Are you in the middle of something or you are not looking at it now?

PR #671, opened 2015-03-16
Implement #369 feature (reportes in HTTP extension).

la10736:
Requester don't need to use busy list. Implemented waiting dictionary and block request
state reset on reset_all requests and polling http errors events.
1. busy information was already in waiting dictionary
2. On reset_all or connection error (polliing error) is better reset all reporter blocks state

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -121,11 +121,19 @@ public class ExtensionManager {
 		result.sortOn('name');
 		return result;
 	}
+	
+	private function resetRequesters(ext:ScratchExtension):void {
+		for(var b:Object in ext.waiting) {
+			delete ext.waiting[b];
+			(b as Block).requestState = 0;
+		}
+	}
 
 	public function stopButtonPressed():* {
 		// Send a reset_all command to all active extensions.
 		for each (var ext:ScratchExtension in enabledExtensions()) {
 			call(ext.name, 'reset_all', []);
+			resetRequesters(ext);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


sclements:
@cwillisf can you take a look at this? we need to make sure this doesn't break Javascript extensions.

src/extensions/ExtensionManager.as
```diff
@@ -442,12 +445,9 @@ public class ExtensionManager {
...
 		function responseHandler(e:Event):void {
-			if(e.type == Event.COMPLETE)
-				b.response = loader.data;
-			else
-				b.response = '';
-
-			b.requestState = 2;
+			if (e.type != Event.COMPLETE){
+				b.requestState = 2;
+			}
```
cwillisf, line 450: Shouldn't this still collect `loader.data`?

src/extensions/ExtensionManager.as
```diff
@@ -162,15 +170,11 @@ public class ExtensionManager {
 		var ext:ScratchExtension = extensionDict[extensionName];
 		if (ext == null) return; // unknown extension
 
-		var index:int = ext.busy.indexOf(id);
```
cwillisf, line 165: I'm nervous about removing the use of `ext.busy` for reporters / requesters. It looks OK at first glance but if possible I'd prefer a more conservative change that only implements HTTP requesters without affecting the way we track requesters in general.

src/extensions/ExtensionManager.as
```diff
@@ -571,6 +572,13 @@ public class ExtensionManager {
 					if (ext.busy.indexOf(id) == -1) ext.busy.push(id);
 				}
 				break;
+			case '_result':
+				i = value.indexOf(' ');
+				if (i == -1) i = value.length;
+				var resid:int = parseInt(value.slice(0, i));
+				var retval:String = value.slice(i + 1);
```
cwillisf, line 579: It may be possible to simplify this by using `value.split` as in the `'_busy'` case.

cwillisf (changes requested):
Sorry this took so long to get to. We're postponing many improvements until Scratch 3.0, but I think this one might make sense to integrate into 2.0 for those still using HTTP extensions.

la10736:
I'll try to take a look to it in the next days... after 1.5 years without working to it is hard to recover all details.

thisandagain:
Thanks for the contribution!

PR #681, opened 2015-03-27
#621 Adding ? At the End Of List Contains Block

### LINKED ISSUE #621: block "list contains [string]" should have question mark at end
the block <[listname] contains [string]> should read <[listiname] contains [string]?> instead. Just add a question mark to be consistent with other Scratch booleans.  

image

2jour:
Adding ? at the end of <list contains> block to make it consistent with other boolean blocks

Fixes #621 

##TestCases : 

1. Make sure <list contains> block works as expected
2. Make sure the other blocks look ok


cwillisf:
:+1:

PR #686, opened 2015-03-30
Remove translation metadata from translator map

cwillisf:
By convention, a PO file contains a metadata entry that uses the empty
string as its key. Our translator was translating the empty string into
this metadata. Now, that entry is removed on dictionary construction.

This fixes LLK/scratch-flash-online#129

### DIFF
--- a/src/translation/Translator.as
+++ b/src/translation/Translator.as
@@ -189,6 +189,7 @@ public class Translator {
 			if (mode == 'val') val += extractQuotedString(line);
 		}
 		if (mode == 'val') dict[key] = val; // recordPairIn(key, val, dict);
+		delete dict['']; // remove the empty-string metadata entry, if present.
 		return dict;
 	}
 


src/translation/Translator.as
```diff
@@ -189,6 +189,7 @@ public class Translator {
 			if (mode == 'val') val += extractQuotedString(line);
 		}
 		if (mode == 'val') dict[key] = val; // recordPairIn(key, val, dict);
+		delete dict['']; // remove the empty-string metadata entry, if present.
```
cwillisf, line 192: I could have prevented the entry from ever making it into the dictionary, but I would have had to put the key check in both places -- there's one inside the loop as well. this seemed more readable and arguably more efficient. But mostly more readable :)

rschamp:
LGTM

2jour:
Looks good on scratch.ly .. :)

Issue #685, opened 2015-03-30
[3D] Unattached Look Block Affects Colour Of Sprite


2jour:
server: production 2.0.97

os: mac osx 10.10
Browser: Safari

1. go to https://scratch.mit.edu/projects/55030264/
2. Click See Inside
3. Click green flag
4. Click Odd Glitch Sprite and change unattached set effect block from ghost to colour. 

Stamper sprite turns grey...

TheLogFather:
I suspect this is related in some way to https://github.com/LLK/scratch-flash/issues/570

sclements:
@TheLogFather I don't think that's it. This project is not using the brightness effect.

The project is stamping a 60% ghosted black rectangle on the stage in an infinite loop. I think that with 3d turned on, which happens when a color effect block is present, the 60% opacity black stage is letting the white background behind it come through and make grey. When 3d is off, I think that the pen layer bitmap probably doesn't have an alpha channel and it will never blend with what's behind it.

Perhaps https://github.com/LLK/scratch-flash/blob/master/src/render3d/DisplayObjectContainerIn3D.as#L938 should be more like:
```
textures.push(new ScratchTextureBitmap(512, 512, false));
```

sclements:
Actually, I was wrong about that. It seems that the stamps are coming out differently in the pen layer itself.

TheLogFather:
Yup, no brightness effect - hence only 'related in some way'...

As for the white background coming through, well, it's exactly the same with a black background... :/

The key piece of info here is that it only happens for colours that are not fully saturated. Note that the rect costume is not actually quite black.

Try it with anything non-grey from the standard colour-picker square (i.e. the 8x7 colour squares), and it all works the same whether or not 3d is enabled. (The quickest way to switch is using ctrl-M, or cmd-shift-M on Mac, which you can do while the project is running.)

But if you try with anything grey, or with another colour that's not fully saturated from the more graded colour-picker, then you'll see the difference.

Hence why I reckon it's 'related' to the difference in the colour-space for the 3d player...

sclements:
It would be nice to have access to call updateEffects() (https://github.com/LLK/scratch-flash/blob/master/src/scratch/ScratchObj.as#L264) so that the cachedBitmap isn't cleared. For something doing a lot of stamping this will cause a performance hit.

PR #687, opened 2015-04-01
Removed unused methods in interface. Cleaned up and optimized stamping.

sclements:
This is a continuation of the fix made for https://github.com/LLK/scratch-flash/issues/685.


cwillisf:
:+1:

PR #690, opened 2015-04-03
Now detecting high resolution bitmap backdrops

### LINKED ISSUE #672: Some backdrop costumes get cut off
If you add a new backdrop costume from the media library and select the baseball-field it looks OK at first. Looking in the costume editor you can see that it's cut off and if you switch costumes and come back to it then it will be cut off in the player as well:
image

sclements:
This fixes #672

### DIFF
--- a/src/ui/media/MediaLibrary.as
+++ b/src/ui/media/MediaLibrary.as
@@ -297,9 +297,9 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
 					if (entry.tags is Array) entry.category = entry.tags[0];
 					var info:Array = entry.info as Array;
 					if (info) {
-						if ((entry.type == 'backdrop') || (assetType == 'costume')) {
-//							entry.width = info[0];
-//							entry.height = info[1];
+						if (entry.type == 'backdrop') {
+							entry.width = info[0];
+							entry.height = info[1];
 						}
 						if (entry.type == 'sound') {
 							entry.seconds = info[0];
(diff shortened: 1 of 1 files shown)


src/ui/media/MediaLibrary.as
```diff
@@ -422,20 +422,26 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
...
 						centerX: item.dbObj.info[0],
 						centerY: item.dbObj.info[1],
 						bitmapResolution: 1
 					};
 					if (item.dbObj.info.length == 3)
 						obj.bitmapResolution = item.dbObj.info[2];
+
+					io.fetchImage(md5AndExt, item.dbObj.name, 0, whenDone, obj);
+				} else { // assetType == backdrop
+					if (item.dbObj.info.length == 2 && item.dbObj.info[0] == 960 && item.dbObj.info[1] == 720)
```
cwillisf, line 443: I would use a >= comparison on the dimensions, so if we somehow end up with an even larger image it still uses high-resolution mode.

sclements, line 443: Since we're so engrained with 480x360 and even high-resolution bitmaps have been abandoned I don't think that will be necessary (at least in this codebase).

cwillisf:
:+1:

2jour:
@shane for project 1000003326 on my port,  Backdrop sunset beach does not cover the whole stage expected? I see this also in production.

PR #691, opened 2015-04-03
Fix pen touches

### LINKED ISSUE #482: Sometimes the pen doesn't draw
http://scratch.mit.edu/discuss/topic/51605/

We should draw a tiny circle instead of a tiny diagonal line:
https://github.com/LLK/scratch-flash/blob/master/src/primitives/MotionAndPenPrims.as#L230

sclements:
When penDown is called, draw a tiny circle instead of a diagonal line. Fixes #482.


cwillisf:
:+1:

PR #698, opened 2015-04-05
Fix integer processing

### LINKED ISSUE #582: Zero point zero isn't zero
Try subtracting zero from 0.0 (e.g. put into a say block).
Or try ceiling of (0.0). :/

See this topic: http://scratch.mit.edu/discuss/topic/79402/

Also, see this project: http://scratch.mit.edu/projects/36859670/

sclements:
This removes the code that added epsilon (1 / 4294967296) to integer block arguments that contained a decimal point. This caused issues with 'ceiling' and comparing values (0.0 != 0).
There were ways to implement this that would not perform well or would increase memory usage. I think I found a good way to fix this while maintaining performance and memory usage.
Fixes #582

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -457,7 +457,7 @@ public class Block extends Sprite {
 		for (i = 0; i < labelsAndArgs.length; i++) {
 			item = labelsAndArgs[i];
 			item.y = indentTop + ((maxH - item.height) / 2) + vOffset;
-			if ((item is BlockArg) && (!BlockArg(item).isNumber)) item.y += 1;
+			if ((item is BlockArg) && (!BlockArg(item).numberType)) item.y += 1;
 		}
 
 		if ([' ', '', 'o'].indexOf(type) >= 0) x = Math.max(x, minCommandWidth); // minimum width for command blocks
@@ -986,10 +986,10 @@ public class Block extends Sprite {
 			space = true;
 			var ba:BlockArg, b:Block, tf:TextField;
 			if ((ba = x as BlockArg)) {
-				s += ba.isNumber ? "(" : "[";
+				s += ba.numberType ? "(" : "[";
 				s += ba.argValue;
 				if (!ba.isEditable) s += " v";
-				s += ba.isNumber ? ")" : "]";
+				s += ba.numberType ? ")" : "]";
 			} else if ((b = x as Block)) {
 				s += b.getSummary();
 			} else if ((tf = x as TextField)) {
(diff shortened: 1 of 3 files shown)


cpseager:
@sclements

Currently the decimal separator was/is still assumed to be the English '.' in a few places

135: 	if (numberType) field.restrict = '0-9e.\\-';
229:         numberType = (field.text.indexOf('.') >= 0 && n is int) ? NT_INT : NT_FLOAT;

However this is not locale safe, as, for instance, German users will want to type 123,45 (comma as decimal separator)

Using NumberFormatter may help this issue:
http://help.adobe.com/en_US/as3/dev/WS9b644acd4ebe59998b99a90125fc4fecfb-7ffb.html

e.g.

//Find the locale decimalSeparator
var nf:NumberFormatter = new NumberFormatter(LocaleID.DEFAULT); 
var decimalSeparator = nf.decimalSeparator;

TheLogFather:
Nice to see this getting fixed in such a straightforward way :)

Not sure some of the logic for the tests are correct though...?

TheLogFather:
Just remembered I can leave comments on the relevant lines - so I'll go do that...

sclements:
@cpseager I would love to support localized numbers. It's a much bigger change than this pull request though. Do we have an issue open for this already?

TheLogFather:
@sclements:
> "The design is such that adding '.0' to both arguments of the 'pick random' block will give you random integers "

Huh, I'm a bit confused...

I believe the behaviour at the moment is that adding the ".0" (to either of the 'pick random' args) signals the intent to get *non-int* values, even when both values are effectively ints.

i.e. "pick random (0.0) to (1)" gives a float in range zero to 1, rather than just either of ints zero or 1.

sclements:
@TheLogFather that was embarrassing but I think the logic is proper now.

TheLogFather:
Ah, good - glad I wasn't just going a bit crazy (not quite yet, anyway...)
As a mathematician, having 0.0 != 0 did feel slightly awkward... ;)

cwillisf:
I'm just glad that ceiling(0.0) won't be 1 any more!
:+1:

TheLogFather:
Looks like this is still broken, as per referenced issue above #711...
-It needs more careful logic for the case when an input is a reporter, rather than something typed.

Even when checking that the inputs are BlockArgs before doing that numberType check, it still needs to deal with the case when one (or both) isn't a BlockArg (i.e. it's a reporter), and check if the resulting values are ints - otherwise it'll always give floating randoms.

TheLogFather:
I think this needs a reference to https://github.com/LLK/scratch-flash/issues/582, so it's a bit more clear what this is fixing...

PR #699, opened 2015-04-05
Fix video transparency in 3D mode

### LINKED ISSUE #679: Video Transparency Does Not Work in 3D Mode
Server: Production
OS: mac osx 10.10
Browser: Safari

1. Go to this project https://scratch.mit.edu/projects/54575222/
2. Video transparency block works
3. Add change colour effect block and video transparency does not work

Expected: Video transparency should work regardless if editor in 3d mode or not

sclements:
Add a tiny number to the stage's ghost effect when the video is on and semi-transparent. This ensures that a shader with ghosting (alpha altering) is used by the stage.
Fixes #679

### DIFF
--- a/src/scratch/ScratchStage.as
+++ b/src/scratch/ScratchStage.as
@@ -23,6 +23,8 @@
 // A Scratch stage object. Supports a drawing surface for the pen commands.
 
 package scratch {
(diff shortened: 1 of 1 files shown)


cwillisf:
:+1:

PR #700, opened 2015-04-06
Extension save load fixes

### LINKED ISSUE #500: Remove extension blocks
When you click on an extension name such as 'WeDo' a 'remove extension blocks' menu appears.

If you click this menu the extension blocks are removed as expected. However it is then not possible to 'Add an extension' once again, for instance in the WeDo example nothing happens when the WeDo image is clicked. 

i.e. once an extension has been 'removed' it is not possible to add that same extension back in again

### LINKED ISSUE #514: .url for extension loaded from .sb2 file
In extensionManager.as:

loadSavedExtensions()

does not make any reference to ext.url, so the extension's

"About <name> Extension" 

drop-down menu does not work after opening an .sb2 file that contains an extension.

However loading a .s2e template directly works ok as loadRawExtension() does correctly process .url

### LINKED ISSUE #557: [3D] JavaScript Extensions not working in offline player
If you try to open a project that contains a JavaScript extension in the offline editor (v429) the extension is not visible under "More Blocks"

It worked in the offline editor v404b, the project would open and the extension blocks were still visible under "More Blocks". After updating to v429 the Extension would not load.

To test, download my [sample project](http://scratch.mit.edu/projects/37255682/) and try to open with the offline editor. Note that the extension blocks are not visible under the "More Blocks" tab.

This was originally reported [here](http://scratch.mit.edu/discuss/topic/61492/)

cpseager:
Fixes for #500, #514, #557 and a few other similar extension issues

1) Fixes #500
Upon new project wedo/picoboard are preloaded (so their icons are available for the Extension Library).
Clicking the icon when selecting the extension simply sets .showBlocks to true. 

However when the 'remove extension blocks' menu was clicked their entry was deleted completely from extensionDict.
This means that the process of 
Load (setEnabled(true)) - Remove (setEnabled(false)) - Load (setEnabled(true)) 
failed on the second load, as the extensionDict entry no longer existed (even though the icon still shown in Library).

Fixed by checking for isInternal before delete

2) 
http extensions were previously never deleted upon 'Remove' menu

3) Fixes #514
ext.url etc. were not saved in extensionsToSave

4) 
ext.port is always present (even if 0), so removed 'else' for 'javascriptURL' in extensionsToSave

5) Fixes #557, #514
loadRawData and loadExtensionData now share code to correct issues like .url missing upon load of an .sb2 file

6) 
[rR] v [r] fix is now done in both directions, as a saved .sb2 file could actually contain either

7)
Removed unused (historical)
ScratchExtension.useScratchPrimitives
ScratchExtension.id

8) Moved extension indicator LED left by 10 pixels so no longer partially obscured by vertical scrollbar (when 2 or more extensions used)

9) Re-instated checkbox for [rR] reporter blocks in the offline version

@sclements for  is 9) ok for [rR] variant?

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -206,9 +206,7 @@ public class ExtensionManager {
 		return ext;
 	}
 			
-			
 	public function loadExtensionData(ext:ScratchExtension, extObj:Object):void {		
-		
 		extensionDict[extObj.extensionName] = ext;
 		ext.port = extObj.extensionPort;
 		ext.blockSpecs = extObj.blockSpecs;
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -75,9 +75,8 @@ public class ExtensionManager {
 	public function specForCmd(op:String):Array {
 		// Return a command spec array for the given operation or null.
 		for each (var ext:ScratchExtension in extensionDict) {
-			var prefix:String = ext.useScratchPrimitives ? '' : (ext.name + '.');
 			for each (var spec:Array in ext.blockSpecs) {
-				if ((spec.length > 2) && ((prefix + spec[2]) == op)) {
+				if ((spec.length > 2) && ((ext.name + '.' + spec[2]) == op)) {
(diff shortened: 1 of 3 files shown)


cpseager:
split into smaller separate pull requests for clarity

PR #702, opened 2015-04-07
Refactor imagespart scriptspart

sayamindu:
This makes it easier to inherit from these classes in derivative builds. Also, I noticed that there are some problems with overriding the ScratchStage() implementation uniformly - this fixes that as well (565810a7d085522da6b0ff8b6b0504fc81606cb9).


sclements:
:+1:

PR #717, opened 2015-04-15
Aborting block drag operations by pressing ESC

AndersBillLinden:
This change will add the function of aborting a drag operation by pressing ESC. That will restore the position/insertion of the block to the place it were when the drag operation was initiated.


AndersBillLinden:
Oh, wrong branch

PR #718, opened 2015-04-15
Aborting block drag operations by pressing ESC.

AndersBillLinden:
This change will add the function of aborting a drag operation by pressing ESC. That will restore the position/insertion of the block to the place it were when the drag operation was initiated.


Scimonster:
I think this is a useful change. It is also unlikely to mess with anyone's
workflow <https://xkcd.com/1172/>, as it currently doesn't do anything.​

AndersBillLinden:
Yes, this is a lifesaver and should be complemented with undo.

sclements:
:+1:

PR #722, opened 2015-04-16
No longer creating an array for every primCall

sclements:
This should reduce some garbage collection for projects with lots of procedure calls.


sclements:
This had a bug.

PR #723, opened 2015-04-16
Rename "presentation mode" to "full screen mode".

AndersBillLinden:
The reason is that the code names this mode in two different ways without an apparent reason. This change requires translation of the string "can drag in fullscreen:" and removes the need of translating the string "can drag in player:".

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -627,8 +627,8 @@ public class Scratch extends Sprite {
 
 	protected var wasEditing:Boolean;
 
-	public function setPresentationMode(enterPresentation:Boolean):void {
-		if (enterPresentation) {
+	public function setFullScreenMode(enterFullScreen:Boolean):void {
+		if (enterFullScreen) {
 			wasEditing = editMode;
 			if (wasEditing) {
 				setEditMode(false);
@@ -641,7 +641,7 @@ public class Scratch extends Sprite {
 			}
 		}
 		if (isOffline) {
-			stage.displayState = enterPresentation ? StageDisplayState.FULL_SCREEN_INTERACTIVE : StageDisplayState.NORMAL;
+			stage.displayState = enterFullScreen ? StageDisplayState.FULL_SCREEN_INTERACTIVE : StageDisplayState.NORMAL;
 		}
 		for each (var o:ScratchObj in stagePane.allObjects()) o.applyFilters();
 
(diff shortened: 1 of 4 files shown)


src/ui/parts/SpriteInfoPart.as
```diff
@@ -75,13 +75,13 @@ public class SpriteInfoPart extends UIPart implements DragClient {
 	}
 
 	public static function strings():Array {
-		return ['direction:', 'rotation style:', 'can drag in player:', 'show:'];
+		return ['direction:', 'rotation style:', 'can drag in fullscreen:', 'show:'];
```
sclements: this shouldn't actually be changed. 'player' is anything but editor mode and doesn't have to be full screen..

sclements:
This is a nice readability improvement to the fullscreen logic. I made a suggestion which affects both 'can drag in player' changes.

src/ui/parts/SpriteInfoPart.as
AndersBillLinden: The text "can drag in player" makes me think that when I check that box, I will not be able to drag the sprite. Instead that is only true in full screen mode.

AndersBillLinden:
Seems like `public function setFullScreenMode(enterFullScreen:Boolean):void {` is a nice improvement over `public function setPresentationMode(enterPresentation:Boolean):void {`

AndersBillLinden:
`stage.displayState = enterPresentation ? StageDisplayState.FULL_SCREEN_INTERACTIVE : StageDisplayState.NORMAL;` was changed to `stage.displayState = enterFullScreen ? StageDisplayState.FULL_SCREEN_INTERACTIVE : StageDisplayState.NORMAL;` which made me think that the change is corrrect.

src/ui/parts/SpriteInfoPart.as
sclements: When checked, the sprite will be draggable in the player when not in full screen mode.

AndersBillLinden:
Another thing that made me think the change is correct is that if you left presentation mode with the setPresentationMode function, edit mode was set to true. If it makes sense to have both false, edit mode should not be set to true there. If my conclusion is correct, we will always be in either one of edit mode and presentation mode.

AndersBillLinden:
@sclements 
````
I made a suggestion which affects both 'can drag in player' changes.
````

where?

src/ui/parts/SpriteInfoPart.as
sclements: If you would change the text back to 'player' in this file I would merge the PR.

AndersBillLinden: How will the user understand what the "player" is?

AndersBillLinden: So what is the "player"?

AndersBillLinden:
Would anyone know what the word "player" relates to in the sentence "can drag in player"? Isn´t the word "fullscreen" more accurate?

AndersBillLinden:
I also wonder if it wouldn´t make more sense to lock the dragging of a sprite in both fullscreen and non fullscreen using that checkbox.

ntlrsk:
I see how "full-screen" is clearer when in the editor, but the phrase "player" comes from the project page in the website. The project is not in full-screen mode, but plays the project with that behavior.

sclements:
:+1:

thisandagain:
Thanks for the contribution!

Issue #727, opened 2015-04-18
C blocks try to snap to hidden blocks


djdolphin:
When you're dragging a C block, it tries to snap to blocks even if they're obscured by the palette:
![screen shot 2015-04-17 at 10 17 34 pm](https://cloud.githubusercontent.com/assets/5834370/7213520/cba7a1ca-e54f-11e4-9d32-fb01e7188aeb.png)
The block is just deleted if you drop it. [I have a video here.](https://e50e8a3cc18b78b96ec6d26e1c13e1c49b4298a4.googledrive.com/host/0Bwg2KFU5SrIoOVF5QUQtZWpUQnc/C%20Block%20Glitch.ogg)

cwillisf:
Fixed by #732

PR #732, opened 2015-04-19
Fixing issue 727

AndersBillLinden:
When dragging a block from the scripts palette to the scripts panel, insertion/wrapping previews will only occur when the mouse is inside the scripts panel.

This issue is reported in https://github.com/LLK/scratch-flash/issues/727

### DIFF
--- a/src/uiwidgets/ScriptsPane.as
+++ b/src/uiwidgets/ScriptsPane.as
@@ -139,11 +139,29 @@ public class ScriptsPane extends ScrollFrameContents {
 	}
 
 	public function updateFeedbackFor(b:Block):void {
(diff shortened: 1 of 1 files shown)


sclements:
Already fixed in #736.

cwillisf:
Looks like a good fix! One style point: could I ask that you put the new nested functions `updateHeight` and `updateFeedbackShape` at the top of `updateFeedbackFor` instead of the bottom?
Thanks for the fix, @AndersBillLinden!

AndersBillLinden:
@cwillisf :+1:

cwillisf:
:+1:
Thanks again for the fix, @AndersBillLinden!

Issue #730, opened 2015-04-18
Colour-picked white no longer draws circle at "pen down"


TheLogFather:
Having 0xFFFFFFFF in the "set pen color to" block doesn't work when just drawing the circle from "pen down" (i.e. then "pen up" without moving the pen).

Anything else, including 0xFFFFFF or even 0xFEFFFFFF, works fine.

If you try to pick white using the eye-dropper, then it doesn't work - I'd assume that's because it's actually putting the value 0xFFFFFFFF in there.

Example project: https://scratch.mit.edu/projects/57657390/

AndersBillLinden:
Maybe this change will not be accepted but the alpha part will instead be cut away.

2jour:
@sclements does merge #736 fix this issue?

sclements:
Yes.

PR #734, opened 2015-04-19
Fixing issue 730

AndersBillLinden:
Use a separate argument to beginFill for alpha.

This will fix the issue reported in https://github.com/LLK/scratch-flash/issues/730


AndersBillLinden:
Maybe, semi opacity should not be added so I close this.

PR #735, opened 2015-04-20
Alternative fix of issue 730, removing alpha component from pen color

AndersBillLinden:
This change just removes the alpha component when setting pen color.

Would fix the nconsistency issued in https://github.com/LLK/scratch-flash/issues/730


sclements:
I implemented an alternative fix which adds the ability to set the alpha as @TheLogFather was attempting to do in his Scratch project. We're going to discuss and decide whether to add this feature or simply fix the bug and not support alpha (as your PR does).

sclements:
Already fixed in #736

PR #736, opened 2015-04-21
Added support for pen alpha

sclements:
Now supporting pen alpha when the pen color format is ARGB and A (alpha) is greater than zero. This means you can have a semi-transparent pen stroke but you cannot have an invisible pen stroke. This is to provide backward compatibility to projects setting the pen color but not the alpha value.

Addresses #730

### DIFF
--- a/src/primitives/MotionAndPenPrims.as
+++ b/src/primitives/MotionAndPenPrims.as
@@ -240,7 +240,9 @@ public class MotionAndPenPrims {
 	private function touch(s:ScratchSprite, x:Number, y:Number):void {
 		var g:Graphics = app.stagePane.newPenStrokes.graphics;
 		g.lineStyle();
-		g.beginFill(s.penColorCache);
+		var alpha:Number = (0xFF & (s.penColorCache >> 24)) / 0xFF;
+		if (alpha == 0) alpha = 1;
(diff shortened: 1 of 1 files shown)


sclements:
Some results using https://scratch.mit.edu/projects/57657390/ from @TheLogFather:
image
And now using alpha:
image

cwillisf:
:+1: on the code, but holding the merge until we can discuss compatibility as a group

TheLogFather:
Wow - wasn't expecting that!
I'd wondered if it was so simple, but I was a bit concerned about how it would behave when drawing with pen alpha on top of previously drawn pen...
Does it draw each new stroke/circle with the alpha over the previous drawing? Or does it draw into a 'layer', so completely overwriting previous drawing with the alpha only getting applied against the backdrop?
If the first, then I'd expect that to be quite hard to use well, and not so useful. It would mean, for example, that drawing a single line would produce an unsightly 'blob' at the start of the line - because there were two things drawn there (the circle at "pen down" followed by the stroke for the line).
I'm hoping it's the second, which, while not ideal for certain things (e.g. trying to draw a translucent surface over a previously drawn surface, for example), is actually usable for a great many things...

PR #739, opened 2015-04-23
Removed logic for handling items from the backpack

sclements:
This causes issues in the offline editor because fromBackpack doesn't exist on MediaInfo objects.


cwillisf:
:+1:

Issue #742, opened 2015-04-29
Long blocks cause a shift left whilst dragging


cpseager:
If you select a long block, such as the WeDo 'set motor direction' and drag it **slowly** to the right into the script area, all of the other blocks suddenly incorrectly shift left (see image). As there is no horizontal scroll bar you can't shift them back again without clicking onto a different palette and then back again..

![scratch_shift](https://cloud.githubusercontent.com/assets/7423187/7393496/19f49772-ee86-11e4-8baa-e14a27e0f682.JPG)

Scimonster:
I've noticed this bug too. Confirmed.

On Wednesday, April 29, 2015, scraxe <notifications@github.com> wrote:

> If you select a long block, such as the WeDo 'set motor direction' and
> drag it *slowly* to the right into the script area, all of the other
> blocks suddenly incorrectly shift left (see image). As there is no
> horizontal scroll bar you can't shift them back again without clicking onto
> a different palette and then back again..
>
> [image: scratch_shift]
> <https://cloud.githubusercontent.com/assets/7423187/7393496/19f49772-ee86-11e4-8baa-e14a27e0f682.JPG>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/LLK/scratch-flash/issues/742>.
>


-- 
Sent from Gmail Mobile

AndersBillLinden:
I notice this bug frequently. My students as well.

TheLogFather:
I'll just drop a note in here to say that allowing this pane to scroll is the currently the only way to see which custom block is which, if you have several long ones that start with the same string (apart from actually grabbing each of them, one by one, until you find the one you want).

Also, clicking on another palette is not the only way to 'unscroll' the pane – there are two other ways...

The simplest is to hold down the shift key and use the scroll-wheel. Unfortunately, this doesn't work on all browsers. The alternative is to hold a block over near the left edge of the pane, thus causing it to scroll back in the opposite direction.

Neither are particularly great, of course!

What would be preferable, I think, would be to allow the divider to be dragged left/right, so you can change the width to whatever suits you at the time. We can drop scrolling at that point...

cwillisf:
Resolved by #744

PR #744, opened 2015-05-03
Fixing issue 742

AndersBillLinden:
Not letting scrolling of ScriptsPane change when carrying a block over it. https://github.com/LLK/scratch-flash/issues/742

### DIFF
--- a/src/util/GestureHandler.as
+++ b/src/util/GestureHandler.as
@@ -133,10 +133,12 @@ public class GestureHandler {
 			}
 		}
 		if (carriedObj && scrollTarget && (getTimer() - scrollStartTime) > SCROLL_MSECS && (scrollXVelocity || scrollYVelocity)) {
-			scrollTarget.contents.x = Math.min(0, Math.max(-scrollTarget.maxScrollH(), scrollTarget.contents.x + scrollXVelocity));
-			scrollTarget.contents.y = Math.min(0, Math.max(-scrollTarget.maxScrollV(), scrollTarget.contents.y + scrollYVelocity));
-			scrollTarget.constrainScroll();
-			scrollTarget.updateScrollbars();
+			if (!scrollTarget is ScriptsPane) {
+				scrollTarget.contents.x = Math.min(0, Math.max(-scrollTarget.maxScrollH(), scrollTarget.contents.x + scrollXVelocity));
+				scrollTarget.contents.y = Math.min(0, Math.max(-scrollTarget.maxScrollV(), scrollTarget.contents.y + scrollYVelocity));
+				scrollTarget.constrainScroll();
+				scrollTarget.updateScrollbars();
+			}
 			var b:Block = carriedObj as Block;
 			if (b) {
 				app.scriptsPane.findTargetsFor(b);


thisandagain:
Resolves GH-742
Resolves GH-1080

cwillisf (approved):

TheLogFather:
Oh, err... Not sure I'm happy about this...

I presume this means we can no longer scroll the pane to see the right side of long blocks? (See https://github.com/LLK/scratch-flash/issues/1080.)

Can we find a temporary alternative to this patch? Maybe have the block palette still scroll when hovering over left/right edges, but have it 'snap back' automatically when the block gets dropped? (Another temporary possibility might be a left/right scrollbar at the bottom?)

cwillisf:
@TheLogFather this change just makes it so that dragging a block near the edge of the block palette won't accidentally scroll the palette area. People often do that accidentally (#742) and "get stuck" because they don't realize how it happened or how to scroll back.

Things like shift+scroll wheel still scroll the palette, but that's only used by a few advanced users like yourself. I agree that we need a better solution like being able to move the divider between the palette and the scripting area or some smarter scrolling system, but that's a larger change than just this one. I suggest using #1080 as a place to talk about that part of the problem.

Note that I've left #1080 open since I don't think this change sufficiently addresses that problem, though I agree with @AndersBillLinden that it fixes #742.

TheLogFather:
Yes, I understood the problem (accidental scroll while hovering at edge during drag – it causes a 'surprise', and such potentially disconcerting behaviour is best avoided), and that the patch only affects drag.

However, shift-scroll appears to work only on certain browsers (works on FF, but not Safari for me).

That's why I suggested the 'snap back' as a potential alternative temporary 'band-aid' fix for the accidental scrolling 'surprise' – yes, the scroll 'surprise' still happens during the drag, but its effects are immediately mitigated on drop.

--

This isn't a particularly big issue – it'll just be a little irritating not having quick access to the right side of longer custom blocks in the "More blocks" palette from within Safari. (Then again, maybe there's something I'm missing in Safari that'll make shift-scroll work...? And any reasonably complicated project I would tend to work on offline with my own compiled version of Scratch that includes all the nice search features. –Though that can be a bit tedious if the project has several cloudvars, since it turns them all into normal globals when saving, and I have to then edit the JSON to switch "persistent" back to true... :/ )

liam4:
> However, shift-scroll appears to work only on certain browsers (works on FF, but not Safari for me).

Doesn't work on any browsers for me (OS X 10.11.6) :(

src/util/GestureHandler.as
```diff
@@ -133,10 +133,12 @@ public class GestureHandler {
 			}
 		}
 		if (carriedObj && scrollTarget && (getTimer() - scrollStartTime) > SCROLL_MSECS && (scrollXVelocity || scrollYVelocity)) {
-			scrollTarget.contents.x = Math.min(0, Math.max(-scrollTarget.maxScrollH(), scrollTarget.contents.x + scrollXVelocity));
-			scrollTarget.contents.y = Math.min(0, Math.max(-scrollTarget.maxScrollV(), scrollTarget.contents.y + scrollYVelocity));
-			scrollTarget.constrainScroll();
-			scrollTarget.updateScrollbars();
+			if (!scrollTarget is ScriptsPane) {
```
TheLogFather, line 136: Doesn't "ScriptsPane" refer to the whole of the scripts pane?
So scrolling (by dragging block near edge) no longer works anywhere in the scripts pane in v450...?
See https://scratch.mit.edu/discuss/topic/220438/
(Oops!)

cwillisf: Yep -- looks like you're right. Thanks for bringing this to my attention!
I'll see if I can find a quick fix. If not I'll revert this change for now.

cwillisf: Here's my attempt to fix both #742 without breaking drag-scrolling in the scripting area: #1207

TheLogFather: Nice! - much more generic solution. :)

PR #747, opened 2015-05-05
Add null check and improve load handling

sclements:
Addresses an issue we're seeing with URLs that take you directly into the editor.

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -256,7 +256,9 @@ public class Scratch extends Sprite {
 
 	public function logException(e:Error):void {}
 	public function logMessage(msg:String, extra_data:Object=null):void {}
-	public function loadProjectFailed():void {}
+	public function loadProjectFailed():void {
+		loadInProgress = false;
+	}
 
 	protected function checkFlashVersion():void {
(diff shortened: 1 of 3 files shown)


src/Scratch.as
```diff
@@ -466,6 +468,7 @@ public class Scratch extends Sprite {
 		removeLoadProgressBox();
 		System.gc();
 		if (autostart) runtime.startGreenFlags(true);
+		loadInProgress = false;
```
cwillisf, line 471: Are you confident that it's safe to set this to `false` before `stepRuntime()` calls `runtime.installNewProject()`?

sclements, line 471: It shouldn't happen before. ScratchRuntime::installProject() is the only caller to projectLoaded() and installNewProject() calls installProject. Scratch's projectLoaded() and loadProjectFailed() are now the only places that loadInProgress are set to false. It's cleaner and clearer this way.

cwillisf, line 471: Ah, cool. Thanks for the explanation!

cwillisf:
:+1:

PR #751, opened 2015-05-07
Changed color to hue on the set pen hue block.

sclements:
setPenColorTo and setPenHueTo with no reporters:
image

with reporters (confusing! they're exactly the same.):
image

FIXED:
image


djdolphin:
For consistency, shouldn't changePenHueBy be changed too?

sclements:
Good call @djdolphin

sclements:
We haven't decided what to do with this yet. There were good reasons in the past to not use the word hue.

PR #752, opened 2015-05-07
Call resetPlugin before installing project or loading extensions.

sclements:
Fixes an issue we're seeing with the WeDo in workshops.


cwillisf:
:+1:

PR #753, opened 2015-05-07
Do not permanently delete internal extensions

cpseager:
Internal extensions should only hide/show blocks, not be permanently deleted, as this stops add-remove-add of that extension from working.


sclements:
:+1:

cwillisf:
:+1:

PR #754, opened 2015-05-07
Move extension LED left

cpseager:
Shift extension LED left so not partly obscured by vertical scrollbar with extensions with many commands.


sclements:
:+1:

PR #755, opened 2015-05-07
Give rR reporter type a checkbox

cpseager:


sclements:
Unfortunately, we can't support this right now. It'll take a bit more work to be able to add watchers to synchronous reporters that have been forced into asynchronous mode.

sclements:
@cwillisf knows more about this.

PR #756, opened 2015-05-07
Share extension code

cpseager:
Share code between LoadRaw and LoadFile for extensions.
Add a fixup for rR and r in both directions.


cpseager:
This pr is out of date now following other more recent extension work.

Issue #767, opened 2015-05-20
Can't import some WAV files


sclements:
Format reference: http://www-mmsp.ece.mcgill.ca/documents/audioformats/wave/wave.html
Example file: http://www.mediafire.com/download/gqimq5471fp5e4e/FNAF+3+Trailer+song.wav

I debugged this a little and found that the file is using the "extended" format chunk data to specify that it's using the IEEE floating point PCM encoding. We don't have code to support that and a cursory search revealed that that most WAV decoders in Actionscript do not support that format either. If we can find an open source one, or do it ourselves, then we can fix this.

Here's my investigation branch: https://github.com/LLK/scratch-flash/tree/IEEE_WAV_decoding

sclements:
I got a little further with my branch but hit a bug in the Flash player:
https://bugbase.adobe.com/index.cfm?event=bug&id=3104536

That method would be perfect for this but I'm not sure if there is a workaround for the bug.

2jour:
64-bit floating point IEEE does not work. Sound turns into a hissing sound after its uploaded.

Also mono does not seem to work for 32 bit floating point IEEE 8000z

1. download wav from  http://mauvecloud.net/sounds/index.html 
2. Upload sound in the editor

2jour:
So just tested for 8000 Hz and here is what i found 

1. 64 bit floating point IEEE 8000Hz works mono and stereo :+1: 
2. 32 bit floating point IEEE 8000z works mono and stereo :+1: 
3. PCM 32 bit does not work (expected?)
4. PCM 8 bit does not work (expected?)
5. u-Law compression (really 'mu'-Law) does not work (expected?)
6. a-law compression does not work (expected?)
7. IMA ADPCM mono works but stereo does not (expected?)
8. MS ADPCM does not work MONO or Stereo (expected?)
9. MPEG Layer 3 works for both Mono and Stereo :+1:

sclements:
I use IMA ADPCM mono internally so the support makes sense.

2jour:
64-bit floating point IEEE does not import on windows 7 chrome 42

2jour:
LGTM

ntlrsk:
I just tried uploading two WAV files that come in but show 0 MB when uploaded to Scratch online. Is this the same or a different issue?

ntlrsk:
Note: I saved as MP3 and that worked.

2jour:
[...]

PR #768, opened 2015-05-20
IEEE WAV audio file decoding

sclements:
We now handle extended WAVE format chunks and encoding type 3 (IEEE Float samples).

Fixes https://github.com/LLK/scratch-flash/issues/767

### DIFF
--- a/src/sound/WAVFile.as
+++ b/src/sound/WAVFile.as
@@ -89,6 +89,8 @@ public class WAVFile {
 				throw Error("WAVFile: can only handle 8-bit or 16-bit uncompressed PCM data");
 			}
 			result.sampleCount = (result.bitsPerSample == 8) ? result.sampleDataSize : result.sampleDataSize / 2;
+		} else if (encoding == 3) {
+			result.sampleCount = Math.floor(result.sampleDataSize / (result.bitsPerSample >>> 3));
 		} else if (encoding == 17) {
 			if (formatChunk.length < 20) throw Error("WAVFile: adpcm format chunk is too small");
 			if (result.channels != 1) throw Error("WAVFile: adpcm supports only one channel (monophonic)");
--- a/src/sound/WAVFile.as
+++ b/src/sound/WAVFile.as
@@ -120,11 +120,18 @@ public class WAVFile {
 	public static function extractSamples(waveData:ByteArray):Vector.<int> {
 		var result:Vector.<int> = new Vector.<int>();
 		var info:Object = WAVFile.decode(waveData);
+		var i:int;
 		var v:int;
 		if (info.encoding == 1) {
 			waveData.position = info.sampleDataStart;
-			for (var i:int = 0; i < info.sampleCount; i++) {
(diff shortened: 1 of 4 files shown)


src/sound/WAVFile.as
```diff
@@ -60,14 +60,22 @@ public class WAVFile {
...
+		result.samplesPerSecond = formatChunk.readUnsignedInt();
+		result.bytesPerSecond = formatChunk.readUnsignedInt();
+		result.blockAlignment = formatChunk.readUnsignedShort();
+		result.bitsPerSample = formatChunk.readUnsignedShort();
+		if (formatChunk.length >= 18 && encoding == 0xFFFE) {
+			var extensionSize:uint = formatChunk.readUnsignedShort();
+			if (extensionSize == 22) {
+				result.validBitsPerSample = formatChunk.readUnsignedShort();
+				result.channelMask = formatChunk.readUnsignedInt();
+				result.encoding = encoding = formatChunk.readUnsignedShort();
```
cwillisf, line 76: Have you tested this `encoding` assignment specifically? Based on the "Extensible Format" section of [this page](http://www-mmsp.ece.mcgill.ca/documents/audioformats/wave/wave.html) it looks to me like you're reading part of a sub-format GUID into `encoding`. Do your docs disagree about what comes after `dwChannelMask`?

src/sound/WAVFile.as
```diff
@@ -109,13 +120,20 @@ public class WAVFile {
...
 			waveData.position = info.sampleDataStart;
-			for (var i:int = 0; i < info.sampleCount; i++) {
+			for (i = 0; i < info.sampleCount; i++) {
 				v = (info.bitsPerSample == 8) ? ((waveData.readUnsignedByte() - 128) << 8) : waveData.readShort();
 				result.push(v);
 			}
+		} else if (info.encoding == 3) {
+			waveData.position = info.sampleDataStart;
+			for (i = 0; i < info.sampleCount; i++) {
+				v = Math.min(32767, Math.max(-32768, waveData.readFloat() * 32768));
```
cwillisf: I would suggest that the innermost expression should be `(waveData.readFloat() + 1) / 2 * 65535 - 32768`. As it stands, the difference between 1.0f and 0.9999f is smaller than I think it should be, and the difference between smaller values is slightly larger than I think it should be. On the other hand, probably nobody will notice ;)

src/sound/WAVFile.as
sclements, line 76: From the spec I linked to: "The first two bytes of the GUID form the sub-code specifying the data format code, e.g. WAVE_FORMAT_PCM. The remaining 14 bytes contain a fixed string, \x00\x00\x00\x00\x10\x00\x80\x00\x00\xAA\x00\x38\x9B\x71."

cwillisf, line 76: Ah, sorry... I missed that you had a link in the issue. Thanks for the explanation... makes sense now :)

cwillisf:
:+1:

PR #780, opened 2015-05-29
Avoid AS3 class properties when caching lower case strings

### LINKED ISSUE #773: [Comparison Operator] constructor < a Returns True
Reporter: https://scratch.mit.edu/discuss/topic/118662/ bugs and glitches
OS: mac osx 10.10.2
Browser: Safari

1. Go to project and play https://scratch.mit.edu/projects/64064376/

Expected: Bat should say false after it says 'constructor < a' . Bat says true because comparison operator returns wrong results

sclements:
Fixes #773

### DIFF
--- a/src/primitives/Primitives.as
+++ b/src/primitives/Primitives.as
@@ -151,16 +151,17 @@ public class Primitives {
 		// This is static so it can be used by the list "contains" primitive.
 		var n1:Number = Interpreter.asNumber(a1);
 		var n2:Number = Interpreter.asNumber(a2);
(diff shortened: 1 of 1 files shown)


cwillisf:
:+1:

griffpatch:
Would I be wrong in saying that "" > 0 no longer evaluates the same either? I'm trying to diagnose some issues I'm having in existing projects, and I think that a null variable (empty string) used to return false for "">0, but now it returns true? Might be worth checking too... please...

PR #792, opened 2015-06-05
Honor resolution setting if present in JSON

### LINKED ISSUE #672: Some backdrop costumes get cut off

cwillisf:
The backdrop JSON can optionally specify the bitmap resolution to use
for a given backdrop. Previously this setting ws ignored on backdrops
and would even prevent the detection of a high-resolution backdrop. Now
it is honored if present, and if absent then the automatic resolution
detection still works.
Fixes #672 (again).

### DIFF
--- a/src/ui/media/MediaLibrary.as
+++ b/src/ui/media/MediaLibrary.as
@@ -440,8 +440,11 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
 
 					io.fetchImage(md5AndExt, item.dbObj.name, 0, whenDone, obj);
 				} else { // assetType == backdrop
-					if (item.dbObj.info.length == 2 && item.dbObj.info[0] == 960 && item.dbObj.info[1] == 720)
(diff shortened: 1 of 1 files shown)


sclements:
I'm surprised this was still broken. Aren't double resolution backdrops always 960x720?

cwillisf:
The problem shows up best when you have a double-resolution backdrop that
explicitly says it's double-resolution. Because of the `item.dbObj.info.length
== 2` check, such a backdrop wasn't covered by the old logic. Now the first
`if` says "if you tell me what your resolution is then I'll believe you"
and the second is the same as before: "if you don't specify your resolution
and you're 960x720 then I'll assume double-resolution."

On Tue, Jun 30, 2015 at 5:54 PM, Shane M. Clements <notifications@github.com

sclements:
:+1:

PR #794, opened 2015-06-08
Optimising lookup of variable names using an Object as a 'dictionary'

TheLogFather:
Uses an Object for faster lookup of variable names.
(-hope I've not missed anything important!)

### DIFF
--- a/src/scratch/ScratchObj.as
+++ b/src/scratch/ScratchObj.as
@@ -54,6 +54,7 @@ public class ScratchObj extends Sprite {
 	public var objName:String = 'no name';
 	public var isStage:Boolean = false;
 	public var variables:Array = [];
+	public var varDict:Object = {}; // lookup 'dictionary' for variables
 	public var lists:Array = [];
 	public var scripts:Array = [];
 	public var scriptComments:Array = [];
@@ -435,10 +436,7 @@ public class ScratchObj extends Sprite {
 
 	public function ownsVar(varName:String):Boolean {
 		// Return true if this object owns a variable of the given name.
-		for each (var v:Variable in variables) {
-			if (v.name == varName) return true;
-		}
-		return false;
+		return varDict["var_"+varName]!=null;
 	}
 
 	public function hasName(varName:String):Boolean {
@@ -452,6 +450,7 @@ public class ScratchObj extends Sprite {
 		if (v == null) { // not found; create it
 			v = new Variable(varName, 0);
 			variables.push(v);
+			varDict["var_"+varName] = v;
 			Scratch.app.updatePalette(false);
 		}
 		return v;
@@ -460,14 +459,8 @@ public class ScratchObj extends Sprite {
 	public function lookupVar(varName:String):Variable {
 		// Look for variable first in sprite (local), then stage (global).
 		// Return null if not found.
-		var v:Variable;
-		for each (v in variables) {
-			if (v.name == varName) return v;
-		}
-		for each (v in Scratch.app.stagePane.variables) {
-			if (v.name == varName) return v;
-		}
-		return null;
+		var v:Variable = varDict["var_"+varName];
+		return v ? v : Scratch.app.stagePane.varDict["var_"+varName];
 	}
 
 	public function deleteVar(varToDelete:String):void {
@@ -478,6 +471,7 @@ public class ScratchObj extends Sprite {
 					v.watcher.parent.removeChild(v.watcher);
 				}
 				v.watcher = v.value = null;
+				delete varDict["var_"+v.name];
 			}
 			else newVars.push(v);
 		}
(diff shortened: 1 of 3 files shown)


griffpatch:
Super! I've been hoping for some movement around these bits of code for a while :) - Thanks @TheLogFather! I wonder if there's anything similar that can be done for event receiver blocks? It could make quite a difference for projects with many clones. Adding one receiver block to a sprite that has 200 clones of itself causes 201 scans of that clone for receivers of that name... And if that sprite has a lot of blocks in it, then that's a lot of scanning!

TheLogFather:
@griffpatch: Yeah, I see what you mean about the broadcast/receive issue - it's checking through every stack in every object (i.e. all sprites/clones & stage) to find a match for the specific broadcast:
https://github.com/LLK/scratch-flash/blob/3ec5460798e7e11069b4a360e1bdf36ebb730f9e/src/interpreter/Interpreter.as#L582
https://github.com/LLK/scratch-flash/blob/3ec5460798e7e11069b4a360e1bdf36ebb730f9e/src/scratch/ScratchRuntime.as#L883

I think it's even worse for 'edge-triggered' hats (things like "when timer/loudness>value") - it looks to me like it's searching for these through every stack in every sprite/clone/stage *at every frame*:
https://github.com/LLK/scratch-flash/blob/3ec5460798e7e11069b4a360e1bdf36ebb730f9e/src/scratch/ScratchRuntime.as#L89
https://github.com/LLK/scratch-flash/blob/3ec5460798e7e11069b4a360e1bdf36ebb730f9e/src/scratch/ScratchRuntime.as#L334
If I'm right about that, it could do with improving...

Fortunately, custom block definition lookups are cached.

TheLogFather:
@griffpatch:
Hopefully, https://github.com/LLK/scratch-flash/pull/805 will address the broadcast issue raised above.

griffpatch:
I think it's great that you're doing these optimisations, I wonder how likely it is that the Scratch Team will actually pick them up? I'd love to know how these changes might affect some of my large projects with hundreds of hat blocks, clones and variables. So what's next on your list ;) - Have you covered the variable access via sensing blocks? or references to lists?

TheLogFather:
I'm not sure how likely it is the ST will take these optimisations, since each of them tends to touch a number of places in the codebase (though a couple of them touch some places only very lightly).
I'm still waiting for some movement on https://github.com/LLK/scratch-flash/pull/230, which was over a year ago now!
I hope they will at least spark off something, though...

I certainly hope that https://github.com/LLK/scratch-flash/pull/797 ("go to definition") will be added fairly soon, at least in some form (with better wording) - that's quite nicely localised. (Though it's not an optimisation like the others - more a 'human optimisation'.)

Of the speedups, the most important one, in terms of raw block interpretation performance, is https://github.com/LLK/scratch-flash/pull/795 - having that recursive checkBlockingArgs test can seriously slow down Scratch if it's dealing with expressions containing several nested levels of operators. (It's actually descending through the argument tree from every block at every level!)

For example, here's an experiment to try...
Calculate 2^40 in a single "set var to" expression that nests 39 multiplication blocks, compared to a repeat (39) that contains a multiplication of a variable by two.
You'd expect the single long expression to be much quicker than the loop (it only has to reference the variable once, compared to 78x for loop) - yet it takes about twice as long! That's all due to the checkBlockingArgs test... removing that knocks down the nested expression time by a factor of 6 or so.

The other one that really needs sorting out is https://github.com/LLK/scratch-flash/pull/796.
Unfortunately, though, what I've done there is really no good - it gives a very nice speedup a lot of the time, but also sometimes causes much worse performance problems. :/
I can't see a clear way to fix it for all cases at the moment, though I'm working through it here and there to see if I can find some alternative.

I suspect the latest https://github.com/LLK/scratch-flash/pull/805 will only have a decent impact on large projects. I do see a modest improvement on my 3d framework project of ~10%, for example. I suspect your PM would see it too.
I'm pretty sure I've made the correct changes in there so that it will run projects with the same behaviour as currently (in particular, the order that broadcasts are received). But my main concern is that I may have missed a point where it should clear the caches during script editing - meaning the order could get temporarily mixed up. Still, it forces a rebuild of the caches after every time a hat block is moved/created/deleted in the editor, or a variable/list is deleted/renamed/created/shown/hidden, or custom block params edited, so it is likely to be really quite temporary indeed...

As for variable access via sensor blocks, that's improved by https://github.com/LLK/scratch-flash/pull/794 because its *two* variable look-ups (one to check if the sprite owns the named var, one to get its value) are done via the dictionary-like Objects (rather than sweeping through an array looking for matches to the var name).
That look-up still isn't as quick as the direct access of https://github.com/LLK/scratch-flash/pull/801, though.

thisandagain:
Thanks for the contribution!

PR #795, opened 2015-06-09
Remove recursive descent through blocks when checking for extension blocking conditions

TheLogFather:
Not got any experience with extensions, so I don't know if this is all done correctly.
In particular, there are numerous questions I've put into various comments.

However, it knocks off a decent chunk of time from raw block processing, so I'm hoping this will at least form the basis for a method to avoid that recursive checkBlockingArgs call (and I've just realised that function itself is still in there but should be removable now...)

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -81,6 +81,7 @@ public class Block extends Sprite {
 	public var requestState:int = 0;		// 0 - no request made, 1 - awaiting response, 2 - data ready
 	public var response:* = null;
 	public var requestLoader:URLLoader = null;
+	public var blockingCount:int = 0;
 
 	public var nextBlock:Block;
 	public var subStack1:Block;
@@ -1016,5 +1017,50 @@ public class Block extends Sprite {
 	protected static function indent(s:String):String {
 		return s.replace(/^/gm, "    ");
 	}
+
+	/* Blocking */
+
+	public function resetBlockingCount():void {
+		// should be called for a top stack block - resets all blocks in stack
+		resetArgsBlockingCount();
+		if(subStack1) subStack1.resetBlockingCount();
+		if(subStack2) subStack2.resetBlockingCount();
+		if(nextBlock) nextBlock.resetBlockingCount();
+	}
+
+	public function resetArgsBlockingCount():void {
+		// descend through all block's args, resetting to zero
+		blockingCount = 0;
+		var args:Array = this.args;
+		for(var i:uint=0; i<args.length; ++i) {
+			var barg:Block = args[i] as Block;
+			if(barg) barg.resetArgsBlockingCount();
+		}
+	}
+
+	public function increaseBlockingCount():void {
+		// increase blockingCount up the chain
+		// TLF: is it only the top level block that needs changing?
+		var b:Block = this;
+		while (true) {
+			b.blockingCount++;
+			if (b.parent is Block) b = Block(b.parent)
+			else return;
+		}
+		return; // never gets here
+	}
+
+	public function decreaseBlockingCount():void {
+		// decrease blockingCount up the chain
+		// TLF: is it only the top level block that needs changing?
+		var b:Block = this;
+		while (true) {
+			if(b.blockingCount>0) b.blockingCount--; // TLF: just in case...
+			if (b.parent is Block) b = Block(b.parent)
+			else return;
+		}
+		return; // never gets here
+	}
+
 
 }}
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -168,8 +168,14 @@ public class ExtensionManager {
 			for(var b:Object in ext.waiting) {
 				if(ext.waiting[b] == id) {
 					delete ext.waiting[b];
-					(b as Block).response = retval;
-					(b as Block).requestState = 2;
+					b = b as Block;
+					if(b) {  // TLF: we probably don't need this...?
+						b.response = retval;
+						if(b.requestState!=2) {  // TLF: just in case...?
+							b.requestState = 2;
+							b.decreaseBlockingCount();
+						}
+					}
 				}
 			}
 		}
@@ -323,6 +329,9 @@ public class ExtensionManager {
 			if(b.isRequester) {
 				if(b.requestState == 2) {
 					b.requestState = 0;
+					// TLF: should it increase the blocking count here instead...?
+					// If so then it'll need also increasing in resetArgsBlockingCount
+					// for any blocks which are requesters.
 					return b.response;
 				}
 				else if(b.requestState == 0) {
@@ -418,6 +427,8 @@ public class ExtensionManager {
 		if (ext == null) {
 			// unknown extension, skip the block
 			b.requestState = 2;
+			// TLF: never increased blockingCount for this, so don't need to decrease?
+			// (unless we should be increasing it when requestState is set to zero...)
 			return;
 		}
 
@@ -425,7 +436,10 @@ public class ExtensionManager {
 			httpRequest(ext, op, args, b);
 		} else if(Scratch.app.jsEnabled) {
 			// call a JavaScript extension function with the given arguments
-			b.requestState = 1;
+			if(b.requestState!=1) { // TLF: just in case...?
+				b.requestState = 1;
+				b.increaseBlockingCount();  // TLF: should this happen when set to zero instead?
+			}
 			++ext.nextID;
 			ext.busy.push(ext.nextID);
 			ext.waiting[b] = ext.nextID;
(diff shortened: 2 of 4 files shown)


cwillisf:
Hi, @TheLogFather! Thanks for making this change. In general it looks pretty good, but there are a few tweaks that I'd like to suggest. Let me know if you disagree with any of my suggestions and we can discuss in more detail.

I believe the right way to think about `requestState` is as follows. In a future change I'd like to make named values for these -- perhaps `EMPTY`, `WAITING`, and `READY` -- but I don't think that should hold up this change.

`requestState` | Meaning | Attempting to read will...
--- | --- | ---
0 | No value yet / empty. | Cause a request, go to state 1, and yield.
1 | Value requested; waiting. | Yield the thread.
2 | Value ready. | "Consume" the value and go to state 0.

src/blocks/Block.as
```diff
@@ -1017,4 +1018,49 @@ public class Block extends Sprite {
...
+		resetArgsBlockingCount();
+		if(subStack1) subStack1.resetBlockingCount();
+		if(subStack2) subStack2.resetBlockingCount();
+		if(nextBlock) nextBlock.resetBlockingCount();
+	}
+
+	public function resetArgsBlockingCount():void {
+		// descend through all block's args, resetting to zero
+		blockingCount = 0;
+		var args:Array = this.args;
```
cwillisf, line 1034: I'd prefer not to make a local variable that hides a class member like this. In this particular case, I think this function is called seldom enough that copying the property to a local variable is not likely to have a high performance impact, so I would just remove the local variable. If you've checked into the performance and it turns out that I'm wrong -- certainly possible :) -- I would recommend renaming the variable to avoid shadowing the member.

src/blocks/Block.as
```diff
@@ -1017,4 +1018,49 @@ public class Block extends Sprite {
...
+	public function increaseBlockingCount():void {
+		// increase blockingCount up the chain
+		// TLF: is it only the top level block that needs changing?
+		var b:Block = this;
+		while (true) {
+			b.blockingCount++;
+			if (b.parent is Block) b = Block(b.parent)
+			else return;
+		}
+		return; // never gets here
```
cwillisf, line 1050: Not a big deal, but this could be written more succinctly as:

```
while (b) {
    b.blockingCount++;
    b = b.parent as Block;
}
```

Also, the `return` at the bottom isn't necessary.

src/extensions/ExtensionManager.as
```diff
@@ -168,8 +168,14 @@ public class ExtensionManager {
 			for(var b:Object in ext.waiting) {
 				if(ext.waiting[b] == id) {
 					delete ext.waiting[b];
-					(b as Block).response = retval;
-					(b as Block).requestState = 2;
+					b = b as Block;
+					if(b) {  // TLF: we probably don't need this...?
```
cwillisf, line 172: I would skip the `if` here. If it turns out that there's a non-`Block` object in the `waiting` array, then that indicates a fairly serious bug elsewhere in the code. I'd rather be informed through an exception rather than hiding the bug silently.

src/extensions/ExtensionManager.as
```diff
@@ -323,6 +329,9 @@ public class ExtensionManager {
 			if(b.isRequester) {
 				if(b.requestState == 2) {
 					b.requestState = 0;
+					// TLF: should it increase the blocking count here instead...?
+					// If so then it'll need also increasing in resetArgsBlockingCount
+					// for any blocks which are requesters.
```
cwillisf, line 334: Nope, no need to increase the blocking count here.
I see this as consuming the value that's ready. This could be the last time that the requester is read for a while (or ever!) so there's no need to block the stack until someone tries to read from the "empty" requester again.

src/interpreter/Interpreter.as
```diff
@@ -317,7 +318,7 @@ public class Interpreter {
 	}
 
 	// Returns true if the thread needs to yield while data is requested
-	public function checkBlockingArgs(b:Block):Boolean {
+/*	public function checkBlockingArgs(b:Block):Boolean {
```
cwillisf, line 321: I would recommend simply removing this function rather than commenting it. If we need to look at or retrieve the code we can always look into the history in git or on GitHub.

cwillisf:
There are two cases I'm slightly concerned about.

What if a requester is pulled out of one stack and added to another stack while its `requestState` is 1? I think it might be necessary to adjust a block's parent's `blockingCount` (recursively upward) when the block is removed or added from/to a stack. This also seems like a reason to store a block count at each level rather than only at the top.

Also, what if the stop button is pressed while a block's `requestState` is 1? The `blockingCount` will be reset to zero when the stop button is pressed, but then if the request finishes it seems like the `blockingCount` will change to -1. It seems like this could easily happen if an extension has a requester which uses, say, network data to come up with its value. A network delay could provide enough time for the user to press the stop button.

@sclements, if you have time I would welcome your input on this change. I suspect you know this code better than I do, so you may catch something I missed.

TheLogFather:
Wow, so many things...

I don't have time right now to answer all of above since I'm in the middle of a conference (I'm actually in Brazil a.t.m.), but I'd say generally that there were a number of things I was rather unsure about (as you can probably tell from the comment questions). However, I was hoping that the PR would get something moving. :)

It's a while since I did this, and I have a recollection that I started off doing it a somewhat different way, so some of the coding probably has a bit of historical junk in it which needs stripping out in the ways you've suggested. -I'll have a closer look in a bit...

src/blocks/Block.as
TheLogFather, line 1034: Yes, I suspect that's some historical junk - will get rid of it.

src/blocks/Block.as
TheLogFather, line 1050: That's much neater :)

src/extensions/ExtensionManager.as
TheLogFather, line 172: OK, good to know - was unsure if that was a possibility

src/extensions/ExtensionManager.as
TheLogFather, line 334: OK, IIRC it looks like the original code was causing a yield also in the case of zero request state - was that unnecessary, then?

src/interpreter/Interpreter.as
TheLogFather, line 321: Yup, should definitely be gone by the time this is done

TheLogFather:
About the two concerning cases...

Yes, I suspect dragging blocks around needs catching, to propagate a decrease/increase up the stck hierarchy. -Always fun when you can edit code while it's actually running, isn't it...

The blockingCount can't go negative (note the check in decreaseBlockingCount), so that particular stop issue won't arise. However, I did wonder how it was meant to behave when stop clicked - if it's really meant to clear things like this (resetAllBlockingCounts in stopAll), or if it should still allow the request to be handled (since the original code didn't seem to have anything special for stopAll...)

src/extensions/ExtensionManager.as
cwillisf, line 334: Sorry I missed this question!
There's a lot of subtlety around requesters and yielding. It's possible I'm wrong about not needing to increase blocking count here, but I still think it's true.
Specifically, I think:
- Trying to read the value when `requestState==0` should cause a request and a yield, and `requestState` should go to 1.
- Trying to read the value when `requestState==1` should just yield.
- Trying to read the value when `requestState==2` should succeed and not yield, and `requestState` should go to zero.

I think of the yield as being associated with a pending request (basically, `requestState==1`). When the state transitions from 2 to 0 there's no active request, so no need to yield. It might be that a request starts immediately afterward, as in the "`else if(b.requestState == 0) {`" case below. If that happens we should yield at that point, but that's not necessarily the case.

TheLogFather:
I made this speed-test project which demonstrates how this descent through the blocks causes a highly significant performance hit as the nesting level increases:
https://scratch.mit.edu/projects/109683095/

After applying this PR, there's no significant difference in timing for the two scripts.

Furthermore, since the top script (less nested) still has several levels of nested operators, it improves the overall speed even for that one – it ends up about half the time. :)

Apart from the main outstanding issue above about dragged blocks, it'd be great if someone with experience with extensions could test this to check for fundamental problems with its behaviour.

adroitwhiz:
Any update on this?

TheLogFather:
@adroitwhiz: Doesn't look like it. :(

The only potential functional issue that I know about would be when splitting apart a script containing a requesting block while the project is running.
It would only be a temporary problem, though, since green-flag resets everything.

However, I'm also not aware this has been tested with actual requestors, so it's possible all sorts of things could go badly wrong. (But I can't see how anything could go wrong for 'normal' Scratch projects, so it'd only affect a small percentage of projects, and I think most likely only temporarily, as mentioned above.)

There are a number of changes to be made to the PR, as mentioned in comments above, but none of them affect the actual functionality, so I would've preferred to see confirmation that it has been tested, and works ok (or not), before making further changes.

thisandagain:
Thanks for the contribution!

Issue #781, opened 2015-05-30
Really odd costume-switch slow-down after saving and reloading project


TheLogFather:
I've noticed this oddity before, but the project below now shows it really well...

Go to the following project: https://scratch.mit.edu/projects/64857642/
Run the tests, and note the time for the non-empty costumes test - it's a lot slower than the other two (as you might expect, really).

Now do the following:
1) Make a remix of the project;
2) Without leaving the editor, run the tests again - I see the non-empty test run *way* quicker;
3) Now reload the project page (of your remixed project);
4) Run the tests again - I find the non-empty case is now much slower again.

I also find it even happens for the original project if I make some changes to it and save it - the non-empty case suddenly becomes really quick. That is, until I reload the project... :/

I get the same behaviour in both Firefox and Safari (on a MacBook).

griffpatch:
I have had consistent cases where my project will always be really slow
playing and editing (as in dragging blocks is slow) until I go to another
costume and drag one of those scripts, then return to the first and it all
runs and plays fast again... I could exit the first sprite and is scripts
for ages and it would not get any faster until I go to the second sprite
and edit something... Very consistent work around.

TheLogFather:
Just saw this drop in as I was checking the page again! :)

Yes, it looks like you can get the speed back by making some changes to the project. The speedup remains after the project is saved, too.

But as soon as you reload, then it's slow again. :/

TheLogFather:
UPDATE to above...
It looks as though the slightest change brings back the speed - but *only* if the project is allowed to save.
If the project is running, so autosave is prevented, then it doesn't show the speed up until I manually save the project. (And then it's gone again on reloading.)

2jour:
Closing issue . #822 seems to fix issue

Issue #750, opened 2015-05-06
Slowdown When Rotate


2jour:
OS mac osx 10.10.2
Browser : Safari
Reporter: Bugs and Glitches

This project demonstrates that rotation is very slow as opposed to goto
https://scratch.mit.edu/projects/60731652

https://scratch.mit.edu/discuss/topic/112922/

Workaround :  performance is noted to be better when using Stage3D

griffpatch:
The slowdown is just a result of how scratch makes use of flash. Whenever
you change the costume for, resize or rotate a sprite, the sprite is
re-rasterized (converted to a bitmap) at the required resolution to be
drawn to the screen. Complex vectors take much longer because resizing a
bitmap is easy, but a vector needs all it's lines, polygons & fills
redrawing every time. Once the sprite is rasterized, it can be moved around
using goto with little to no expense because it does not need to be
redrawn, just pasted to the screen in a new location (this is called bitmap
caching).

It is worth noting that perhaps the 3d hardware mode using hardware sprite
rotation and keeps the originally cached bitmap unchanged? But that is just
a guess to explain why it would be faster as I have not looked at the 3d
rendering path in scratch.

TheLogFather:
@griffpatch: yes, I thought that too - but the OP suggested in the topic that it wasn't always so slow if there was no touching test happening during the same time.

That made me wonder if there was something going on with caching the bitmap - since it doesn't actually need to cache it if it doesn't have to get all the pixels right for a collision test, or stamp it. (Presumably, Flash just has to stick the object on the screen in the right place, so Scratch doesn't need to make its own bitmap.)

So I went back and checked... and it's definitely slower now!

Something happened after v423g which had a detrimental effect on rotation.
But, looking at the changes for v424, it's really difficult to see what could be causing it... I'm suspicious that the change might actually have been something slightly earlier in github, that is related to bitmaps, after ~423f, but which wasn't really included in the v423g release...?

TheLogFather:
A bit more info...

https://scratch.mit.edu/projects/60472556/ is a good test project to for this, so I did some experimenting.

There's a clear slowdown from v423g to v424 (as retrieved from djdolphin's Scratch archive - should really checkout from github to compare...)

But the crucial data point is that if I add a touching edge test in there then v423g becomes as slow as v424 - which is basically what the OP claimed in the discussion topic...

griffpatch:
Nice work there :)

2jour:
[...]

PR #796, opened 2015-06-10
Calculate new MD5 sums immediately when base/textLayerData change

TheLogFather:
This appears to fix both https://github.com/LLK/scratch-flash/issues/781 and https://github.com/LLK/scratch-flash/issues/750

In the case of https://scratch.mit.edu/projects/64857642/ (referenced in #781 above), it speeds up the non-empty costume case (middle one) by a factor of 10x or so, which is just what is seen immediately after saving the project (because the MD5s are then set correctly...)

I'm not sure I entirely understand how this base/textLayerData setter is meant to work, but it seems like it isn't working as intended with the MD5 sums at the moment because they are not getting set correctly somehow, so the shape is not getting cached except right after saving, which means costume switching is much slower than it could be unless the project was just saved. Also, it affects other things too, as seen in the rotation issue (#750).

### DIFF
--- a/src/scratch/ScratchCostume.as
+++ b/src/scratch/ScratchCostume.as
@@ -110,7 +110,7 @@ public class ScratchCostume {
 
 	public function set baseLayerData(data:ByteArray):void {
 		__baseLayerData = data;
-		baseLayerMD5 = by.blooddy.crypto.MD5.hashBytes(data) + fileExtension(data);
+		baseLayerMD5 = data ? by.blooddy.crypto.MD5.hashBytes(data) + fileExtension(data) : null;
 	}
 
 	public function get textLayerData():ByteArray {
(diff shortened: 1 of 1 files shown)


TheLogFather:
So, after going on a hunt for the introduction of these setters, I discovered https://github.com/LLK/scratch-flash/commit/d4d2c8397dc9c65e219f62d9369ba9f2866bbb48 looks kinda familiar. But then  https://github.com/LLK/scratch-flash/commit/b90b81494938fd6c555cb15a37c40f916fc591a4 happens...

In that case, I don't actually understand what really needs doing to make this work correctly - in fact, I'm not even sure what 'correctly' is...? :/

sclements:
Closed in favor of https://github.com/LLK/scratch-flash/pull/822

PR #797, opened 2015-06-11
Adds new item "go to definition" for drop-down menus of custom block callers

TheLogFather:
Really missed having this! -And so simple to add, too...

Not sure about the best text for it. Currently used "go to definition", but maybe "show definition"?
(It's a pity that, in English, the word 'definition' seems so different to the word 'define', as used on the block definition - makes the link potentially not quite so clear.)

I guess translations will need adding for it...

### DIFF
--- a/src/scratch/BlockMenus.as
+++ b/src/scratch/BlockMenus.as
@@ -534,22 +534,21 @@ public class BlockMenus implements DragClient {
 		addGenericBlockItems(m);
 		m.addItem('edit', editProcSpec);
 		if (block.op==Specs.CALL) {
-			/// TLF: alternative text? -maybe "show definition"...?
-			m.addItem('go to definition', gotoProcSpec);
+			m.addItem('see the define script', jumpToProcDef);
 		}
 		showMenu(m);
 	}
 
-	private function gotoProcSpec():void {
+	private function jumpToProcDef():void {
 		if(!app.editMode) return;
 		if (block.op != Specs.CALL) return;
 		var def:Block = app.viewedObj().lookupProcedure(block.spec);
 		if (!def) return;
 		var pane:ScriptsPane = def.parent as ScriptsPane;
-        if (!pane) return;
+		if (!pane) return;
 		if (pane.parent is ScrollFrame) {
-		   pane.x = 5-def.x*pane.scaleX;
-		   pane.y = 5-def.y*pane.scaleX;
+		   pane.x = 5 - def.x*pane.scaleX;
+		   pane.y = 5 - def.y*pane.scaleX;
 		   (pane.parent as ScrollFrame).constrainScroll();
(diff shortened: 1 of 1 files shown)


griffpatch:
You the man :) please please scratch team take these awesome changes!

ntlrsk:
I see this could be very helpful. For wording how about "see the define script" or "go to define block"? (Since "show" usually means on the Stage).

src/scratch/BlockMenus.as
```diff
@@ -533,9 +533,28 @@ public class BlockMenus implements DragClient {
 		var m:Menu = new Menu(null, 'proc');
 		addGenericBlockItems(m);
 		m.addItem('edit', editProcSpec);
+		if (block.op==Specs.CALL) {
+			/// TLF: alternative text? -maybe "show definition"...?
+			m.addItem('go to definition', gotoProcSpec);
```
sclements: Can you please change this to 'see the define script'?

TheLogFather: Yup, I'll do that a bit later today - that 'define script' was the wording I'd come to prefer myself. :)
I did wonder about whether to merge this with #861 (the block highlighting PR) in some way - maybe, along with the "highlight callers" item (don't really like 'callers' that much...), having a "highlight define script" item, which actually also jumps to the script like this.
But I guess that PR is a considerably larger change, and this one does stand better by itself, so...

TheLogFather: OK, I made the text change, plus a couple of minor tweaks.
My only comment after trying it out as "see the define script" would be that it makes the menu seem a bit wide compared to others (might some translations make it even wider?), and a tad unusual since other menu items (and tooltips) don't really use articles, i.e. "the", "a", etc...
Shall we start up SUREMI (Society for the Use of Real English in Menu Items)... ;)

ealgase:
This seems simple, but it seems that it would be very laggy.

TheLogFather:
@ealgase: why would you think that's the case...?

sclements:
Okay, looks like we're changing our mind about the text. Could you please make it say, "define"?

TheLogFather:
Just the single word "define"...?
I guess that makes a degree of sense - it's sorta like it's a verb (as with many other items), so you're telling the editor you want to 'define' this block... kinda... bit subtle, though...

ntlrsk:
@TheLogFather Yes, we were worrying about translation getting long, so the editor team ended up thinking this would be the most direct wording (even if a little subtle).

TheLogFather:
TBH, I think quite often it's not worth worrying overly much about these things, since we so often find things out by experience rather than by actually understanding the meaning - i.e. once we *see* what something does when we try it, we understand how to use it (and may even largely ignore the meaning behind the words themselves - just consider the ubiquitous "File" menu...)
Anyway, should be ready to push in a mo...

PR #799, opened 2015-06-11
Don't try to decode zero length ByteArrays.

sclements:
Fix for https://github.com/LLK/scratch-flash-online/issues/168


cwillisf:
:+1:

PR #801, opened 2015-06-13
Fast access for variables via direct index in block.

TheLogFather:
Each variable keeps track of its index in its owner's variable list (1-based position, and negative if global, i.e. on stage).
This index is consistent, most of the time, across clones, and so can be used by the block for fast access. This is not true under certain conditions - in particular, when a sprite/clone dynamically creates its own variables while running script, and when the user deletes a variable. These cases are dealt with by setting a zero index for dynamically created vars (so the block doesn't access them directly), and by a new application-wide Boolean 'varsAreDirty', which is set true when a variable is deleted. However, on clicking GF or stop, this boolean is cleared, along with all block index caches which then become ready to be refilled on next run (since there are no longer any clones that can cause inconsistent indexing).

If ever Scratch provides capability for dynamic deletion of variables, this method will become less useful...

In terms of speedup, it brings variable access in line with the speed of custom block param access (see https://scratch.mit.edu/projects/40989674/).
It’s worth testing this change on some of my speedtest projects: https://scratch.mit.edu/studios/795672/ The improvement for raw block-processing (i.e. not things to do with bitmaps, etc.) is around 10-20%.

Together with https://github.com/LLK/scratch-flash/pull/795 and https://github.com/LLK/scratch-flash/pull/796 (if that can be sorted out correctly), there's a substantial speed increase all around.

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -96,6 +96,7 @@ public class Scratch extends Sprite {
 	protected var wasEdited:Boolean; // true if the project was edited and autosaved
 	private var _usesUserNameBlock:Boolean = false;
 	protected var languageChanged:Boolean; // set when language changed
+	public var varsAreDirty:Boolean = true;  // cleaned when stop/GF clicked
 
 	// UI Elements
 	public var playerBG:Shape;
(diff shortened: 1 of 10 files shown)


thisandagain:
Thanks for the contribution!

PR #804, opened 2015-06-16
Update Primitives.as

togazo:
for example... when n=14 (etc.),
'Math.pow(10, n)' is '100000000000000',
but
'Math.exp(n * Math.LN10))' was '100000000000000.13'...


sclements:
:+1:

PR #805, opened 2015-06-16
Optimisations for hat block processing

TheLogFather:
Hat blocks "when I receive"/"when backdrop switches", "when key pressed", "when this clicked" and "when /somesensor/" each now has its own cache for each original sprite. (Clones do not need a separate cache, since they share scripts with the original sprite. Collecting the relevant hat blocks for a clone just means effectively taking them from the original sprite.)
This avoids sweeping through all scripts of all sprites/clones/stage every time a key press/repeat happens, a broadcast is sent, or something is clicked. Also avoids the same sweep through all scripts of all sprites/clones/stage at every runtime step, looking for 'edge trigger' hat blocks.
Note that the cache for receivers is an Object which allows fast look-up of relevant receivers for a particular message.

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -559,6 +559,7 @@ public class Block extends Sprite {
 			dup.fixExpressionLayout();
 			dup.fixStackLayout();
 		}
+		Scratch.app.runtime.clearAllHatCaches();
 		return dup;
 	}
 
@@ -816,6 +817,7 @@ public class Block extends Sprite {
 		newStack.x = p.x + deltaX;
 		newStack.y = p.y + deltaY;
 		Scratch.app.gh.grabOnMouseUp(newStack);
+		Scratch.app.runtime.clearAllHatCaches();
 	}
 
 	public function deleteStack():Boolean {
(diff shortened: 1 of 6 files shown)


sclements:
I'm swamped but I'll look at your PRs soon.

TheLogFather:
OK, thanks!

Glad to see the md5sum issue sorted (at least, I hope it is - haven't yet had time to actually test it...)

Many of the optimisation PRs involve quite a bit of messing around in various places, and could do with testing on a wide array of different project types to ensure everything behaves correctly - it's so easy to miss something (like the current issue with the parameter index cache, i.e. https://github.com/LLK/scratch-flash/issues/810).

It'd be good to see https://github.com/LLK/scratch-flash/pull/797 in there a.s.a.p., though (but with the modified wording as suggested by ntlrsk), since that's highly localised and really useful.

Also, the simple fixes for https://github.com/LLK/scratch-flash/issues/810 and https://github.com/LLK/scratch-flash/issues/815 I'd class as "obvious". :)

bobbybee:
@TheLogFather Any benchmarks?

TheLogFather:
Parts of this likely need updating to work with various more recent changes in the Scratch source.
(E.g., two that come to mind would be broadcast renaming, and vars in certain sensor hat blocks.)
Somehow, I'd missed bobbybee's question about benchmarks above. I should go back and see if I can take a look at that again. But this is all relatively low priority, since it's only really complex projects, with lots of separate scripts, that would see any improvement from this.

thisandagain:
Thanks for the contribution!

PR #812, opened 2015-06-22
Saved 'host' extension property in project

kurtdevocht:
The 'host' property is currently not saved in the extension section of the project, and it always defaults to '127.0.0.1' when a project is opened, even is the Scratch extension file (.s2e) has onther host defined.

Windows makes a distinction between 'localhost' and '127.0.0.1'. Hosting an http server on '127.0.0.1' is only allowed as an administrator. Hosting an http server on 'localhost' can be done as a regular user.

When building a Scratch http extension, hosting it on 'localhost' makes it easier to use. However, when a project is opened that contains an extension, the host will be reset to '127.0.0.1'. You need to reload the .s2e file to set the host to 'localhost'.

As far as I can see, saving the host property in the Scratch project file only has benefits.

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -139,6 +139,7 @@ public class ExtensionManager {
 			descriptor.extensionName = ext.name;
 			descriptor.blockSpecs = ext.blockSpecs;
 			descriptor.menus = ext.menus;
+			if(ext.host) descriptor.host = ext.host;
 			if(ext.port) descriptor.extensionPort = ext.port;
 			else if(ext.javascriptURL) descriptor.javascriptURL = ext.javascriptURL;
 			result.push(descriptor);


cpseager:
I think the host property is not saved as it is considered a security issue.

The local Windows hosting issue can probably be resolved a different way, by simply changing this default string to 'localhost' instead of 127.0.0.1.

https://github.com/LLK/scratch-flash/blob/master/src/extensions/ScratchExtension.as#L45

kurtdevocht:
Scratch as ddos-attack tool, cool ;-)

Changing the default to 'localhost' might be a solution, but might break existing helper apps too... Maybe its not a bad idea to save the host only if it's 'localhost'?

thisandagain:
Hi @kurtdevocht , we'd prefer not to extend the SB2 format to track the host as per security / privacy concerns.

PR #813, opened 2015-06-22
3d refactor

cwillisf:
These changes improve the performance and (hopefully) reliability of rendering the Scratch stage area using 3D-style rendering calls. The 3D rendering mode is activated for any project which contains effect blocks from the Looks category, or for all projects in the tablet version.

Fixes LLK/scratch-flash-online#125

Things to test for projects which run in 2D mode:
* There should be no noticeable change.

Things to test for projects which run in 3D mode:
* Project graphics look the same as they did before.
* Project performs the same as before or better.
* Sprites on the stage should still be draggable when in the editor.
* Switching between presentation mode, the editor, and the project page should not cause problems.
* Activating the screensaver, minimizing the browser, locking the computer, and otherwise interfering with rendering should not cause problems.


sclements:
:+1:  Yay!

PR #821, opened 2015-06-30
3d cleanup

cwillisf:
I had removed the argument since pre-multiplied alpha lets us use the same blend mode for both stamps and normal drawing, making for simpler code. However, blending is cheaper when the destination factor is zero. With that in mind it may be a better idea to turn off blending when we can, especially on mobile platforms.
Also, disable error checking on Context3D. I left the line in there so it's easy to change it to `true` when debugging.


sclements:
:+1:

PR #822, opened 2015-06-30
Calculate MD5 if it's null

sclements:
Fixes both #781 and #750

### DIFF
--- a/src/scratch/ScratchCostume.as
+++ b/src/scratch/ScratchCostume.as
@@ -257,6 +257,7 @@ public class ScratchCostume {
 
 	private static var shapeDict:Object = {};
 	public function getShape():Shape {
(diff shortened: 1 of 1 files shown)


cwillisf:
:+1:

2jour:
I don't notice a fix for #750 though however i see an improvement for #781

TheLogFather:
@2jour: how are you testing #750?
It looks better to me for https://scratch.mit.edu/projects/60472556/

sclements:
I get 30fps rotating the clones in 3D and 8fps in 2D.

TheLogFather:
Yes, 3D is much quicker.
But isn't the comparison (at least for now) for 2D against v437?
I get ~10fps with 438, but ~3 with 437...

PR #823, opened 2015-06-30
Merged ServerOnline.callServer() features upward

cwillisf:
These are remnants from the ScratchX merge.
- We now try to reload crossdomain.xml in the event of a security error.
- During a whenDone(null) call which is reporting an error,
  Server.callServerErrorInfo contains information about the attempt.


sclements:
:+1:

PR #824, opened 2015-07-01
Fix logo positioning

cwillisf:
Fix logo position problems from ScratchX merge and a build problem for the offline editor


sclements:
:+1:

PR #826, opened 2015-07-02
Added a method for setting default urls that can be overridden.

sclements:
Necessary to fix the online scratch version after merging ScratchX.


cwillisf:
:+1:

PR #827, opened 2015-07-02
Microworlds

sclements:
This adds a new layout mode for the interest based pathways research project. It's not done but it's good enough for the rest of the team to use in their design iterations.


cwillisf:
:+1:

PR #828, opened 2015-07-02
Convert sounds to native Flash Sound objects and use them for playback.

sclements:
This makes Flash play sounds smoothly and independently of project playback issues.

### DIFF
--- a/src/sound/ScratchSoundPlayer.as
+++ b/src/sound/ScratchSoundPlayer.as
@@ -70,43 +70,72 @@ public class ScratchSoundPlayer {
 	public function ScratchSoundPlayer(wavFileData:ByteArray) {
 		getSample = getSample16Uncompressed;
 		if (wavFileData != null) {
-			var info:* = WAVFile.decode(wavFileData);
-			soundData = wavFileData;
-			startOffset = info.sampleDataStart;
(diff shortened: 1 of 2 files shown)


src/sound/ScratchSoundPlayer.as
```diff
@@ -70,43 +70,72 @@ public class ScratchSoundPlayer {
...
+				stepSize = info.samplesPerSecond / 44100.0;
+				if (info.encoding == 17) {
+					adpcmBlockSize = info.adpcmBlockSize;
+					getSample = getSampleADPCM;
+				} else {
+					if (info.bitsPerSample == 8) getSample = getSample8Uncompressed;
+					if (info.bitsPerSample == 16) getSample = getSample16Uncompressed;
+				}
 			}
+			catch (e:*) {}
```
cwillisf: Should we log this exception?

cwillisf:
:+1:

src/sound/ScratchSoundPlayer.as
sclements: I added the try/catch because of an exception I saw in Sentry about the waveFileData being length zero. Now it's logging just in case there are other exceptions.

PR #832, opened 2015-07-06
Added workaround for Flash Sound bug.

sclements:
This fixes a sound bug present in Flash 11-14 (https://bugbase.adobe.com/index.cfm?event=bug&id=3104536)


sclements:
Remaining issues:
Can we release original sound memory? not if they edit the sound
NotePlayer needs to play nicely with the new method. Perhaps we should create and cache drum and note sounds. It would be nice if their creation didn't slow down the project playback. How can we best handle this?

PR #833, opened 2015-07-06
Revert sound refactor

sclements:
We found a few issues and there isn't enough time to resolve them before the deploy.


cwillisf:
:+1:

PR #834, opened 2015-07-06
Override protocols in URL overrides

cwillisf:
If the SWF is loaded over HTTP or HTTPS then overrideURLs will replace the protocol specified by the incoming URLs with a protocol matching the SWF's URL. If the SWF was not loaded over HTTP or HTTPS then overrideURLs will not modify the incoming URLs.
This is effectively the same behavior that the online editor used before the ScratchX merge.


sclements:
:+1:

Issue #836, opened 2015-07-07
[3D] Fire Jumper Does Not Work


2jour:
OS windows 7
Browser: chrome 43

1. Go to staging project 1000005029 and you will notice block falls through the line
2. Go to https://scratch.mit.edu/projects/33696538/ and it behaves differently

PR #841, opened 2015-07-08
Bring back texture filtering workaround

cwillisf:
The recent 3D refactor broke the `forcePixelate` flag, which used to force the pixelate effect on under certain conditions. Using the "touching color _?" block with thin lines depended on this feature: linear filtering often causes the thin line to be blended too much with surrounding elements, resulting in failed detection.

This change is similar in spirit to the `forcePixelate` flag from before, but uses texture filtering instead of the pixelate effect. This should result in improved performance when the pixelate effect is not in use.

This fixes LLK/scratch-flash#836

### DIFF
--- a/src/render3d/shaders/fragment.agal
+++ b/src/render3d/shaders/fragment.agal
@@ -119,7 +119,7 @@ ft0.xyzw *= Tex_wh
 ft0.xy += Tex_u0v0
 
 // Get the texture pixel using ft0.xy as the coordinates
-ft1 = tex<2d,clamp,linear,nomip>(ft0, fs0)
+ft1 = tex<2d>(ft0, fs0)
 
 /*** ft1 == (r, g, b, a) ***/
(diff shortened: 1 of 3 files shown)


src/render3d/shaders/fragment.agal
```diff
@@ -119,7 +119,7 @@ ft0.xyzw *= Tex_wh
 ft0.xy += Tex_u0v0
 
 // Get the texture pixel using ft0.xy as the coordinates
-ft1 = tex<2d,clamp,linear,nomip>(ft0, fs0)
+ft1 = tex<2d>(ft0, fs0)
```
cwillisf, line 122: I removed these parameters to avoid confusion: they will be ignored now that we're using `setSamplerStateAt()`

sclements:
Weird that it's not auto-building. I wonder if Travis isn't using the right playerglobal file (Context3DTextureFilter exists in 11.6).

Issue #837, opened 2015-07-07
[3D] Video is 0% Transparent


2jour:
Version: Staging 438
OS : Windows 7
Browser: Chrome 43

1. Go to staging project 1000005035 .. You will notice 0% transparency
2. https://scratch.mit.edu/projects/63809044/
3. Also go to https://scratch.mit.edu/projects/69725610 This also does not have 0% transparency like in project 2

Expected: To see the backdrop when video is on like production

Issue #838, opened 2015-07-07
[3D] Can't Change Video Transparency


2jour:
Version staging v438
Browser: chrome 43
OS : windows 7

Please see staging project 1000005036
vs 

https://scratch.mit.edu/projects/54575222/

cwillisf:
Fixed by #846

PR #846, opened 2015-07-13
Apply video transparency as ghost effect

cwillisf:
The previous method of making the video image transparent relied on the fact that all objects were drawn with the same shader, and allowed the video image's alpha value to be used in place of the ghost effect. This broke when we started determining the shader on a per-object basis. Now video transparency is implemented by applying the standard ghost effect to the bitmap object representing the video image.

This fixes LLK/scratch-flash#837 and LLK/scratch-flash#838
See issues for test cases


sclements:
:+1:

Issue #845, opened 2015-07-09
[3D] Presentation Mode Shrinks Item On Stage


2jour:
OS: Windows 7
Server: Staging v438
Browser: chrome 43

1. Go to the following project in staging and play in presentation mode 1000005072
 You will see this 
![image](https://cloud.githubusercontent.com/assets/6530036/8602848/96e64340-2642-11e5-84e2-cd646af1cd4b.png)

2. Ctrl +m and you will see the following
![image](https://cloud.githubusercontent.com/assets/6530036/8602889/d50be97c-2642-11e5-8b58-8c8f8f85e170.png)

This is not the case in production

https://scratch.mit.edu/projects/69959724

cwillisf:
Fixed by #847

PR #847, opened 2015-07-13
Ensure correct appScale when activating 3D mode

cwillisf:
Instead of calling `setRenderView()` when setting up the 3D context, call `onStageResize()` which calls `setRenderView()` after calculating `appScale` and collecting the UI transform.

Test case: see issue #845


sclements:
:+1:

Issue #857, opened 2015-07-16
String Value true and Infinity not working


2jour:
https://scratch.mit.edu/discuss/topic/134166/

TheLogFather:
See my latest comment in https://github.com/LLK/scratch-flash/pull/780

2jour:
https://github.com/LLK/scratch-flash/pull/780#issuecomment-121958892

PR #859, opened 2015-07-16
Fix argument comparisons

sclements:
Fixes https://github.com/LLK/scratch-flash/issues/857
I tested 1/0 == 'Infinity' (true) and '' > 0 (false)


cwillisf:
:+1:

Issue #815, opened 2015-06-24
Hacked "set/change/show/hide var" block causes exception when renaming any var


TheLogFather:
Dunno if you want to bother fixing this since it's only a 'hacked' block issue, but I thought I'd point it out anyway (and it's a simple fix)...

The exception happens in ScratchRuntime.as when it's searching for references to the var name, and it comes across the offending 'set/change/show/hide' block, and assumes the first arg will have an argValue (rather than an expression that would normally get evaluated).

The line of exception in release v437 (since it's now changed in master since original post):
https://github.com/LLK/scratch-flash/blob/release/v437/src/scratch/ScratchRuntime.as#L834

The consequence is that renaming a variable doesn't change references in the scripts (and renaming appears to not even happen at first, since it doesn't do the palette update straight away).

TheLogFather:
I put this (and the (_of_) exception, #831) into a project so you could more easily see what I mean:
https://scratch.mit.edu/projects/70022728/  (not bothered to share it)

cwillisf:
Fixed by @TheLogFather with #949

PR #861, opened 2015-07-17
Adds block and sprite (in sprite pane) highlighting capability.

TheLogFather:
This extends the current concept used with the broadcast/receiver menu items "show senders/receivers" to provide a more general capability to highlight sprites that contain certain things, as well as even the individual blocks within those sprites.

Three examples are included (see pics in next comment for demos):

1) extends the "show senders/receivers" menu items (renamed "highlight senders/receivers") to also highlight the relevant blocks in the highlighted sprite ("clear senders/receivers" also renames to the standard "clear highlights", used for all other highlights too);

2) adds item "highlight variable" to variable getter menus (along with the standard "clear highlights") which highlights the sprites that use the variable (only current sprite, if it's a local var), and all the blocks in those sprites which use it ("set", "change", "hide/show" and getters);

3) adds menu item "highlight callers" to custom block definition and callers (suggestion for better text?) which highlights the current sprite (since no global defines so far!) and all the uses of the custom block.

It's not hard to think of other cases where it would be useful to highlight which sprites use certain things, and where those blocks are in the scripts...

(Note this PR also fixes some exceptions caused by hacked blocks, such as https://github.com/LLK/scratch-flash/issues/815)
Sorry it's only against v437 (hence conflicts) - started this a little while back, only just got around to finishing it off, but hopefully it'll give the concept reasonably well...

Hope that all makes sense! -Any questions, let me know.


TheLogFather:
Huh... the broadcast one seems to have broken somehow (can't be just 'cos of last minute fixes to tabs...?)

Gotta shoot off , so not got time to debug it right now - will come back to it in a bit and commit a fix.

TheLogFather:
Here are a couple of screenshots I took that show examples of the highlighting...

1) Right-click on the variable gives the usual menu (shown near centre), but now with the "highlight variable" item. After clicking that item it highlights, in the sprite pane, the sprites that use the (global) variable, as well as all blocks within the sprites. (Highlighting is preserved when switching sprites.)
highlight-var-demo

2) Right-click on a caller (or definition) gives the usual menu (shown middle-bottom), but now with the "highlight callers" item (could do with better name?) Highlights callers in sprite, as shown.
highlight-demo-callers

Could be a useful tool, I think...

griffpatch:
Love it

Thepuzzlegame:
Would **love** to see this implemented!

TheLogFather:
A couple more demo pics to go with the newer commits, showing the generic "highlight this block" menu item for a parameter block and an operator block.
highlight-demo-params

Note in the second case that it's designed to highlight only those operator blocks "[function v] of ( )" that have the same function as the current block - I suspect that'd be the preferred way to do it for that one (rather than highlighting every occurrence with other functions)...?
highlight-demo-operator

I've also made variable blocks, list blocks, switch costume blocks, and numerous other blocks, work in this kind of way (so they only highlight matching blocks which also have the same var/list/costume/etc.)
Some of them make sense (the above operator blocks, stop blocks, etc.), but some may not be the best choice. However, they're there since it's easier to remove them than to add them.

src/scratch/BlockMenus.as
```diff
@@ -509,6 +509,11 @@ public class BlockMenus implements DragClient {
 		}
 		m.addItem('help', block.showHelp);
 		m.addLine();
+		if (highlightItems) {
+			m.addItem('highlight this block',genericHighlightSameBlocks); // "highlight same blocks", "highlight all like this"...?
```
TheLogFather: Or "highlight these"? Or even just "highlight"...?

Thepuzzlegame:
Perhaps change the highlight color from 0xaeff00 to 0xE0E000 so as to be the same color as highlighted sprites and the stage. Also the blur is a little too much for me, maybe bump it down from 4 to 3?

TheLogFather:
@Thepuzzlegame: I chose to make it not yellow for various reasons. One being that it gets a bit lost for blocks that are within yellow control blocks. Also, to distinguish it from the glow around running scripts. And it's also more similar in *appearance* to the sprite pane highlight than the values suggest - especially when it 'mixes' with the blue frame for the selected sprite/stage. (But I still think I'll tweak that one slightly to make it match the block highlighting - rather than tweaking the block highlighting colour.)

As for the blur level, I find it's maybe a tad strong on default size. But it also starts to get a little feeble when zooming in (i.e. making blocks look bigger). I've been considering making it having a degree of dependence on the script pane scaling, so I'll give it a go and see how it looks...

TheLogFather:
BTW, @Thepuzzlegame, I meant to ask whether you're just looking at the pics above or if you've actually compiled this and tried running it yourself?
The reason I ask is because the pics have the scripts view zoomed out one level, which makes the blocks smaller compared to the blur - for normal script zoom level you may have a different opinion about the amount of blur...

Thepuzzlegame:
@TheLogFather Now that you explain it I get your reasoning for not making the block highlight yellow (although I do think it is a good idea to tweak the sprite/stage highlighting to match the block highlighting). I had previously gone ahead and compiled and run this and the blur at normal level was a *little* too blurry for me. As you've said, consider looking in to having the blur level change with the script pane scaling. I think your idea for the "<" and ">" buttons is great :)

What do you think of having in the right click menu for the scripts panel the option to clear highlights?

Loading image...

TheLogFather:
Screenshot showing highlight widget (near bottom-right):
highlight-demo-widget
The widget only appears when there are highlighted blocks in the currently viewed sprite/stage. The central circular part performs the 'clear highlight' function (and so also makes it disappear).
Note that this commit also makes sprites remember their last script pane position - meaning that when you select a sprite in the sprite pane, the scripts pane will be scrolled to the same place it was last left (rather than being reset).

Thepuzzlegame:
@TheLogFather Looking good!

TheLogFather:
Thanks - I think it's coming together.

There are a couple of fairly major changes I need to make to it, though...

1) Sort out the order of highlights. At the moment, it goes through the blocks in the stacking order, which means that, although it goes through highlighted blocks within a stack from top to bottom as you'd expect, it goes through different stacks in a way that seems quite haphazard. I'm planning to make it go through the stacks in more of a top-to-bottom then left-to-right kind of order in the scripts pane, which would feel better when cycling through the highlighted blocks, I think.

2) Merge it all into v438 - more specifically, into the version with the much better refactored block-rendering branch merged. It looks like that will affect a number of things here, so there's some work to do for that...

Thepuzzlegame:
Well, good luck ;)

TheLogFather:
Well, that's (1) basically sorted (though the coding could do with some tidying up, I think - there are probably better ways to achieve some of the things it does, but I've got no previous experience of using actionscript, so I'm kinda making things up a bit here!)

Next job is to update to v439 and see what needs changing to fit with that...

liam4:
This is awesome and I hope to see it implemented in the future. :)

TheLogFather:
I've been putting off working on this further until there was movement on the block render refactor, since that would have a pretty major impact on how this all works.
But if that's now closed then I'm not clear if I should now proceed further with this, probably clean it up here and there, take some feedback/ideas for changes and/or a bit more to add to it, etc...?

jwzimmer:
@TheLogFather, Thank you so much for working on this feature! The documentation and planning are really great - the screenshots showing how the feature works, the extremely thoughtful comments and conversation - just awesome work! : )
However, we're not going to add it to scratch-flash.
As we focus on Scratch 3.0, we're focusing on developing new features there, and we will consider adding this feature to our next major release. 
We really appreciate that you have taken the time to contribute to Scratch! The scratch-blocks and scratch-vm repositories are great places for open source contributions right now... Hope to see you there. : )
Thanks again!

liam4:
Btw, it's already partially implemented in Scratch 3.0:

Demo

See the "Glows" controls in the demo.

TheLogFather:
Yeah, had a feeling that a number of PRs would be heading in that direction after the S3 announcement.

Ah well...

Issue #798, opened 2015-06-11
[Categories Block] [Performance] Slow to Switch Categories When Green Flag Is Clicked


2jour:
Reporter: https://scratch.mit.edu/discuss/topic/121042/?page=1#post-1071738

OS: mac osx 10.10.2
Browser Safari

1. Go to this project https://scratch.mit.edu/projects/66488314/#editor 
2. Click green flag
3. Try to switch categories like Looks or Motion

Behaviour: Appears  like you can't switch categories. However, sometimes the switch does occur after a considerable time

Expected : Should be able to switch categories even with green flag clicked

Marking high because not sure if workaround is obvious to user: workaround click stop

sclements:
When testing this project I was unable to change the block category at all. I brought it into the debugger and found that the show/hide variable blocks [cause a full refresh](https://github.com/LLK/scratch-flash/blob/master/src/scratch/ScratchRuntime.as#L979) of the palette categories and current palette. This refresh prevents the user from being able to change the category. Implementing code to explicitly toggle the watcher checkboxes instead of refreshing would resolve this.

sclements:
Also discussed here: https://scratch.mit.edu/discuss/topic/134434/

PR #862, opened 2015-07-17
Optimize toggling watcher visibility

sclements:
Don't update the palette if nothing changed. Fixes https://github.com/LLK/scratch-flash/issues/798.


cwillisf:
:+1:

PR #864, opened 2015-07-21
Moved any key in menu

sclements:
Puts the 'any' before the letter keys and after the arrow keys. Moves 'space' to the top.


cwillisf:
:+1:

PR #866, opened 2015-07-23
Don't thrown an exception when loadBytes fails.

sclements:
Fixes https://github.com/LLK/scratch-flash-online/issues/168


cwillisf:
:+1:

PR #868, opened 2015-07-23
Block render refactor

### LINKED ISSUE #848: Projects with large (huge) amounts of scripts crash Flash (possibly freezing the browser) on loading
I suspect this has to do with the GUI trying to render the scripts, so I'm not sure whether this is a bug in Scratch or a result of my carelessness. Thoughts?

[ This fix is by no means urgent, as I can always find ways to shrink the scripts significantly, but this is still not the intended behaviour. ]

sclements:
This improves performance of rendering large stacks of blocks and fixes an issue where very large stacks can crash the editor. Fixes #848.

It's been tested by @2jour, @bobbybee, and @TheLogFather.

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -536,13 +536,7 @@ public class Block extends Sprite {
 	public function removeBlock(b:Block):void {
 		if (b.parent is BlockStack) (b.parent as BlockStack).removeBlocks(b);
 		else if (b.parent == this) removeChild(b);
-		if (b == nextBlock) {
-			var rb:Block = b;
-			if (parent)
-				while (rb = rb.nextBlock)
-					parent.removeChild(rb);
-			nextBlock = null;
-		}
+		if (b == nextBlock) nextBlock = null;
 		if (b == subStack1) subStack1 = null;
 		if (b == subStack2) subStack2 = null;
 		if (b.isReporter) {
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -537,6 +537,10 @@ public class Block extends Sprite {
 		if (b.parent is BlockStack) b.parent.removeChild(b);
 		else if (b.parent == this) removeChild(b);
 		if (b == nextBlock) {
+			var rb:Block = b;
+			if (parent)
+				while (rb = rb.nextBlock)
+					parent.removeChild(rb);
 			nextBlock = null;
 		}
 		if (b == subStack1) subStack1 = null;
@@ -776,7 +780,13 @@ public class Block extends Sprite {
 		}
 		// TODO: Remove any waiting reporter data in the Scratch.app.extensionManager
 		if (parent is Block) Block(parent).removeBlock(this);
-		else if (parent) parent.removeChild(this);
+		else if (prevBlock) prevBlock.removeBlock(this);
+		else if (parent) {
+			var bs:BlockStack = parent as BlockStack;
+			parent.removeChild(this);
+			if (bs && !bs.firstBlock && bs.parent)
+				bs.parent.removeChild(bs);
+		}
 		this.cacheAsBitmap = false;
 		// set position for undelete
 		x = top.x;
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -550,7 +550,7 @@ public class Block extends Sprite {
 	public function insertBlock(b:Block):void {
 		var oldNext:Block = nextBlock;
 
-		if (oldNext != null) parent.removeChild(oldNext);
+		if (oldNext != null && oldNext.parent == parent) parent.removeChild(oldNext);
 
 		if (parent) parent.addChild(b);
 		b.prevBlock = this;
@@ -564,7 +564,7 @@ public class Block extends Sprite {
 		b.x = this.x;
 		b.y = this.y - b.height + BlockShape.NotchDepth;
 		parent.addChild(b);
-		(parent as BlockStack).firstBlock = b;
+		(parent as BlockStack).setFirstBlock(b);
 		b.bottomBlock().insertBlock(this);
 	}
 
@@ -573,7 +573,7 @@ public class Block extends Sprite {
 		b.y = this.y - b.base.substack1y(); //  + BlockShape.NotchDepth;
 		parent.addChild(b);
 		if (parent is BlockStack)
-			(parent as BlockStack).firstBlock = b;
+			(parent as BlockStack).setFirstBlock(b);
 		b.subStack1 = this;
 		this.prevBlock = b;
 		b.fixStackLayout();
@@ -796,7 +796,8 @@ public class Block extends Sprite {
 	public function objToGrab(evt:MouseEvent):* {
 		if (isEmbeddedParameter() || isInPalette()) return new BlockStack(duplicate(false, Scratch.app.viewedObj() is ScratchStage));
 		// TODO: Implement grabbing blocks from within a stack and support canceling the drag (reverting to original position, parent, etc)
-		return (parent is BlockStack ? parent : this);
+		 if (parent is BlockStack && (parent as BlockStack).firstBlock == this) return parent;
+		return this;
 	}
 
 	/* Events */
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -534,7 +534,7 @@ public class Block extends Sprite {
 	}
 
 	public function removeBlock(b:Block):void {
-		if (b.parent is BlockStack) b.parent.removeChild(b);
+		if (b.parent is BlockStack) (b.parent as BlockStack).removeBlocks(b);
 		else if (b.parent == this) removeChild(b);
 		if (b == nextBlock) {
 			var rb:Block = b;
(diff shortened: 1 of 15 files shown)


src/blocks/Block.as
```diff
@@ -550,9 +464,18 @@ public class Block extends Sprite {
...
-		if (nextBlock != null) dup.addChild(dup.nextBlock = nextBlock.duplicate(forClone, forStage));
-		if (subStack1 != null) dup.addChild(dup.subStack1 = subStack1.duplicate(forClone, forStage));
-		if (subStack2 != null) dup.addChild(dup.subStack2 = subStack2.duplicate(forClone, forStage));
+		if (nextBlock != null) {
+			dup.nextBlock = nextBlock.duplicate(forClone, forStage);
+			dup.nextBlock.prevBlock = dup;
+		}
+		if (subStack1 != null) {
+			dup.subStack1 = subStack1.duplicate(forClone, forStage);
+			dup.subStack1.prevBlock = dup;
```
cwillisf, line 473: Should this be (or include) `dup.subStack1.parent = dup`?
Edit: I see other places where you use `prevBlock` to point from the first block of a substack to the block containing that substack. How do you tell the difference between this and a simple stack?

src/blocks/Block.as
```diff
@@ -727,14 +669,18 @@ public class Block extends Sprite {
 	}
 
 	public function topBlock():Block {
-		var result:DisplayObject = this;
-		while (result.parent is Block) result = result.parent;
-		return Block(result);
+		var result:Block = this;
+		while (result.parent is Block)
+			result = result.parent as Block;
```
cwillisf, line 668: Does this reliably climb out of substacks in the new model?

sclements, line 668: I believe this is logically the same but I switched DisplayObject to Block which makes it stricter and matches the return value. This code will crawl up nested arguments if called on a reporter Block. The final logic will utilize the BlockStack if present to get the top block reference.

src/blocks/Block.as
sclements, line 473: I did this to have non-DOM references between the Blocks. When the user switches to view the duplicate sprites blocks they will get added to the ScriptPane within BlockStack containers. At that point they will be added to the DOM.

src/blocks/Block.as
```diff
@@ -833,7 +780,13 @@ public class Block extends Sprite {
 		}
 		// TODO: Remove any waiting reporter data in the Scratch.app.extensionManager
 		if (parent is Block) Block(parent).removeBlock(this);
-		else if (parent) parent.removeChild(this);
+		else if (prevBlock) prevBlock.removeBlock(this);
+		else if (parent) {
```
cwillisf, line 778: I think the `if (parent)` and `if (prevBlock)` checks here might need to be reversed

src/blocks/Block.as
```diff
@@ -863,14 +816,16 @@ public class Block extends Sprite {
...
 		if (scriptsPane) scriptsPane.addComment(this);
 	}
 
 	/* Dragging */
 
-	public function objToGrab(evt:MouseEvent):Block {
-		if (isEmbeddedParameter() || isInPalette()) return duplicate(false, Scratch.app.viewedObj() is ScratchStage);
+	public function objToGrab(evt:MouseEvent):* {
+		if (isEmbeddedParameter() || isInPalette()) return new BlockStack(duplicate(false, Scratch.app.viewedObj() is ScratchStage));
+		// TODO: Implement grabbing blocks from within a stack and support canceling the drag (reverting to original position, parent, etc)
```
cwillisf, line 821: stale TODO

src/render3d/DisplayObjectContainerIn3D.as
```diff
@@ -1463,24 +1465,4 @@ public class DisplayObjectContainerIn3D extends Sprite {SCRATCH::allow3d{
 		return uiContainer;
 	}
 }}
-}
-
-import flash.utils.getQualifiedClassName;
-
-internal final class Dbg {
```
cwillisf, line 1470: :joy:

src/scratch/ScratchRuntime.as
```diff
@@ -346,7 +348,7 @@ public class ScratchRuntime {
 
 			SCRATCH::allow3d {
 				// Should we go 3D?
-				if(isGraphicEffectBlock(b))
+				if(!app.isIn3D && isGraphicEffectBlock(b))
```
cwillisf, line 351: This new check isn't actually necessary: go3d() checks internally. It won't hurt, though.

src/uiwidgets/ScriptsPane.as
```diff
@@ -535,51 +550,51 @@ return true; // xxx disable this check for now; it was causing confusion at Scra
...
-			for each (var c:Array in columns) {
+			for each (var c:Vector.<BlockStack> in columns) {
 				if (fitsInColumn(b, c)) {
 					assigned = true;
 					c.push(b);
 					break;
 				}
 			}
-			if (!assigned) columns.push([b]); // create a new column for this stack
+			if (!assigned) columns.push(Vector.<BlockStack>([b])); // create a new column for this stack
```
cwillisf, line 578: I believe `new <BlockStack>[b]` avoids creating an array before the vector

cwillisf:
:+1:

bobbybee:
What's up with the Travis errors?

cwillisf:
@bobbybee thanks for noticing!

We started having trouble with our Travis builds a few weeks ago when I made a change in some of the 3D rendering code. The code was fine -- it was a problem with the way that we set up our Travis builds. I fixed this in Master a few days ago. These are the errors displayed on the "push" build here: https://travis-ci.org/LLK/scratch-flash/builds/73969819

However, it looks like the failures on the "pr" build might be something different. Do you know what's going on with these errors, @sclements?
https://travis-ci.org/LLK/scratch-flash/builds/73969811

TheLogFather:
I noticed a bug with this a while back, but only just got around to updating my fork of this branch to ensure it's still there in latest (which it is...)
When creating a custom block, it doesn't initially show up in the "More Blocks" pane until its define block is moved again in the scripts pane (at which point it's fine).

TheLogFather:
Another bug I'm seeing is an exception when it tries to hide/show a list or var when something is clicked from the blocks palette (i.e. NOT the scripts pane).

The simplest example is just to click either show or hide of a var while in the "Data" pane:

TypeError: Error #1009: Cannot access a property or method of a null object reference.
	at interpreter::Interpreter/toggleThread()
	at ui::BlockPalette/clear()
	at scratch::PaletteBuilder/showBlocksForCategory()
	at ui::PaletteSelector/select()
	at ui.parts::ScriptsPart/updatePalette()
	at Scratch/updatePalette()
	at scratch::ScratchRuntime/showVarOrListFor()
	at primitives::SensingPrims/primShowWatcher()
	at interpreter::Interpreter/evalCmd()
	at interpreter::Interpreter/stepActiveThread()
	at interpreter::Interpreter/stepThreads()
	at scratch::ScratchRuntime/stepRuntime()
	at Scratch/step()

It also happens if you click a custom block in the blocks palette, and that script shows/hides a var/list. (However #798 'fixes' that, since it no longer changes anything in the blocks palette.)

TheLogFather:
I've been using this for a while now, and I've come across a number of little issues here and there.

I'm somewhat hesitant to bring them up, though, because they are (mostly) transient, and it's not at all clear how to reproduce them. But I thought I'd mention them to see if others come across them, and maybe have better idea how to track down the causes...

1) Script placement can get reset to top-left when opened in 'normal' editor.
This was a major, one-of issue I had yesterday where the scripts all looked fine from within this version of the editor, but when I saved and opened the project from a 'normal' version of the editor, all the scripts of a particular sprite were placed at top-left (and that subsequently crashed the player, both online and standalone). Opening from this version of the editor was fine (all scripts placed correctly).
Unfortunately, after fixing it by various means, I accidentally overwrote the sb2, so I don't have the project.json that was causing it. :(

2) Occasional temporary misplacing of drop target highlights.
When dragging a block, occasionally it won't drop where it should be able to drop - and if I move around I can sometimes see it showing the expected drop target highlights offset by some amount (a hundred pixels or so). After dropping, it usually ends up all in the correct place again.
Next few times it happens, I'll keep a note of how far and which direction the offset is, and see if there's consistency to it.

3) Occasional placement of scripts slightly too far to the left.
This is minor - really only cosmetic - but I've very rarely seen it place a script so it appears as if its leftmost boundary is a few pixels further left than the limits of the script pane - so the leftmost few pixels cannot be scrolled into view.
Only had it happen for a couple of scripts, but ISTR they continued to behave like that even after saving/reloading (until (1) above happened, and then all the script positions got reset...)

TheLogFather:
I think the problems with (2) above arise when prepending blocks to a stack. -It can actually cause a number of cosmetic issues with block rendering, too (e.g. a gap where the second block should be...)

sclements:
Thanks for the testing @TheLogFather ! This is very helpful.

TheLogFather:
Does this mean this improvement is 'dead'?

I've certainly been wondering if there was anything else going on with it, since it's been a while, and it does still have numerous (mainly cosmetic) problems.

But it was a huge improvement in the speed when manipulating long/complicated scripts, so it'd be great to try to revisit it at some point.

PR #871, opened 2015-07-24
Improve framerate to 60fps instead of 30fps

DevanWolf:
It'll be nice if Scratch projects will run better in 60fps rather than the odd 30fps.


bobbybee:
This is significantly more CPU intensive, no?

liam4:
Bump, I suggest an option in the user preferences for this.

bobbybee:
@liam4 Might fragment the community..

sclements:
This change doesn't fit in the current Scratch design.

PR #872, opened 2015-07-24
Fixes to some blocks

DevanWolf:
Fix some bugs to certain blocks, such as some block duplicates and extra ends. This would be more nicer.

### DIFF
--- a/src/Specs.as
+++ b/src/Specs.as
@@ -110,7 +110,7 @@ public class Specs {
 	}
 
 	public static var commands:Array = [
-		// block specification					type, cat, opcode			default args (optional)
+		// block specification					type, cat, opcode				default args (optional)
 		// motion
 		["move %n steps",						" ", 1, "forward:",					10],
 		["turn @turnRight %n degrees",			" ", 1, "turnRight:",				15],
@@ -129,9 +129,9 @@ public class Specs {
 		["set y to %n",							" ", 1, "ypos:",					0],
 		["--"],
 		["if on edge, bounce",					" ", 1, "bounceOffEdge"],
-		["-"],
-		["set rotation style %m.rotationStyle",	" ", 1, "setRotationStyle", 		"left-right"],
 		["--"],
+		["set rotation style to %m.rotationStyle",	" ", 1, "setRotationStyle", 	"left-right"],
+		["-"],
 		["x position",							"r", 1, "xpos"],
 		["y position",							"r", 1, "ypos"],
 		["direction",							"r", 1, "heading"],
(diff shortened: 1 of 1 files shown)


cwillisf:
@DevanWolf I see that there are some changes in here beyond bug fixes... we may need to review the design changes. Could you list the things that you've changed? For example:
* Make backdrop blocks available on sprites, so that any sprite may change the backdrop.
* Blocks with drop-down costume, backdrop, or sound selection now allow users to type the name of a costume in addition to allowing drop-down selection.
* Changed block text from "this" to "that" (etc.)

src/Specs.as
```diff
@@ -355,33 +356,32 @@ public class Specs {
...
+		["%m.list contains %s?",							"b", 12, "list:contains:"],
 		["-"],
 		["show list %m.list",								" ", 12, "showList:"],
 		["hide list %m.list",								" ", 12, "hideList:"],
 
 		// obsolete blocks from Scratch 1.4 that may be used in older projects
-		["play drum %n for %n beats",			" ", 98, "drum:duration:elapsed:from:", 1, 0.25], // Scratch 1.4 MIDI drum
-		["set instrument to %n",				" ", 98, "midiInstrument:", 1],
-		["loud?",								"b", 98, "isLoud"],
+		["play drum %n for %n beats",			" ", 98, "drum:duration:elapsed:from:", 48, 0.25], // Scratch 1.4 MIDI drum
```
cwillisf, line 365: Why 48?

src/Specs.as
```diff
@@ -355,33 +356,32 @@ public class Specs {
...
 
 		// obsolete blocks from Scratch 1.4 that are converted to new forms (so should never appear):
 		["abs %n",								"r", 98, "abs"],
 		["sqrt %n",								"r", 98, "sqrt"],
 		["stop script",							"f", 98, "doReturn"],
 		["stop all",							"f", 98, "stopAll"],
-		["switch to background %m.costume",		" ", 98, "showBackground:", "backdrop1"],
-		["next background",						" ", 98, "nextBackground"],
+		["switch backdrop to %m.costume",		" ", 98, "showBackground:", "backdrop1"],
+		["next backdrop",						" ", 98, "nextBackground"],
```
cwillisf, line 375: These block specs need to stay the way they were: they need to match the specs from Scratch 1.4 in order to import old projects correctly.

src/Specs.as
```diff
@@ -355,33 +356,32 @@ public class Specs {
...
 		// testing and experimental control prims
-		["noop",								"r", 99, "COUNT"],
+		["noop",								"r", 99, "NOOP"],
 		["counter",								"r", 99, "COUNT"],
 		["clear counter",						" ", 99, "CLR_COUNT"],
 		["incr counter",						" ", 99, "INCR_COUNT"],
-		["for each %m.varName in %s",			"c", 99, "doForLoop", "v", 10],
+		["for each %m.var in %n",				"c", 99, "doForLoop", "variable", 10],
 		["while %b",							"c", 99, "doWhile"],
-		["all at once",							"c", 99, "warpSpeed"],
```
cwillisf: Why was this removed?

cwillisf:
If you "git fetch" and then "git rebase origin/master" then that will fix the build error.

PR #879, opened 2015-08-01
Set up Gradle build scripts and tell Travis to use them

cwillisf:


cwillisf:
@sclements I'm going ahead with this merge but you should probably take a look when you get the chance. `scratch.gradle` is intended to also be accessed from other repositories that depend on this one, so it's sort of a library of common functionality, whereas `build.gradle` is specific to this repository.

config.groovy
```diff
@@ -0,0 +1,18 @@
+environments {
```
cwillisf, line 1: The `ConfigSlurper` class requires the `environments` keyword here.

PR #882, opened 2015-08-03
Show Load Experimental Extension only in dev mode

cwillisf:
The button only works on ScratchX so it should be hidden elsewhere.
This fixes LLK/scratch-flash-online#186


sclements:
:+1:

PR #885, opened 2015-08-04
Fix exception when processing `key () pressed?`

### LINKED ISSUE #880: [Regression] [Sensor Key Pressed] Arrow Keys not working consistently
Running on Scratch.ly on version 439 using Chrome on Win 7

 Staging project 35785780

Using the Left/Right/Up/Down arrow keys should move the cat around.  It doesn't work in the editor, nor from the project page, nor in full screen mode.  Oddly enough, it does work while the Ask/Answer window is waiting for text (click on the bug to trigger the Ask/Answer).

This is a pretty big problem as arrow keys are used very widely in game projects.

cwillisf:
When 'any' key support went in, this code was rearranged so that checking `stack.args[0]` was no longer protected by the `stack.op` check through short-circuiting. This change makes the `stack.op` check happen first again, accessing `stack.args` only if that check passes.
This fixes #880

### DIFF
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -191,11 +191,13 @@ public class ScratchRuntime {
 		if (30 == ch) keyName = 'up arrow';
 		if (31 == ch) keyName = 'down arrow';
 		if (32 == ch) keyName = 'space';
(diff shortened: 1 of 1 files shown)


sclements:
:+1:

PR #896, opened 2015-08-20
Make sure the backbuffer isn't too small

sclements:
Fixes an exception we're seeing.
http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display3D/Context3D.html#configureBackBuffer() - "The minimum size of the buffer is 32x32 pixels."


rschamp:
:+1:

PR #897, opened 2015-08-20
Bump down the max texture size.

sclements:
This gives slightly less overall texture memory and uses smaller textures. Currently our top exception is from trying to use textures that are too big and unsupported by the user's system.

The exception is ArgumentError: Error #3683: Texture too big (max is 2048x2048).


rschamp:
:+1:

PR #898, opened 2015-08-20
Don't play beyond the end of our buffer

sclements:
This should resolve an exception we're seeing in playBuffer (RangeError: Error #1125).


rschamp:
LGTM, just curious how it can get into that state?

sclements:
I don't know how it happens. Perhaps we'll have another bug someday for the sound editor where the sound doesn't play correctly. at least now there will be no exception to break the editor.

2jour:
Is this in production @sclements , @cwillisf ?

cwillisf:
It looks to me like this went out with 439.2, yeah.

PR #899, opened 2015-08-20
add null check

sclements:
This should fix an exception we're seeing in ScratchCostume::computeTextLayer (TypeError: Error #2007).


rschamp:
:+1:

bobbybee:
Might this be related to #1047 ?

PR #900, opened 2015-08-20
Use app's stage instead

sclements:
If the instance has been removed from the stage then this.stage will be null. Use the app's stage to avoid a null stage value.


cwillisf:
:+1:

Issue #468, opened 2014-08-20
"when [loudness v] > (reporter block)" hat event block causes the project to not run


DevanWolf:
When you try to run the project with a variable or reporter in a "when [ v] > ()" block, it won't work!
There's something about the way Scratch deals with variables in these event blocks that means it gets confused when a project is first run (curiously, it's fine if you've run the project beforehand *without* the variable in there, and then you put the variable there and run it - my guess would be that something internal is not getting initialised before this event trigger starts getting checked).
Example: http://scratch.mit.edu/projects/20324557

cwillisf:
Fixed by @TheLogFather with #913

PR #901, opened 2015-08-21
Fix error when primVarGet is called without an active thread.

sclements:
This occurs when there is a timer hat block which can run before the green flag is pressed.

### DIFF
--- a/src/interpreter/Interpreter.as
+++ b/src/interpreter/Interpreter.as
@@ -677,6 +677,8 @@ public class Interpreter {
 	// a reference to the Variable object is cached in the target object.
 
 	private function primVarGet(b:Block):* {
+		if (activeThread == null) return 0;
+
 		var v:Variable = activeThread.target.varCache[b.spec];
 		if (v == null) {
 			v = activeThread.target.varCache[b.spec] = activeThread.target.lookupOrCreateVar(b.spec);


TheLogFather:
Ohhhh... will this likely fix https://github.com/LLK/scratch-flash/issues/468?
Shouldn't it at least make some attempt to get the correct value for the var, via some means of identifying the target?

TheLogFather:
...otherwise "when timer/loudness>(var)" could cause some confusion when it always triggers at the start, even though (var) may've been deliberately saved with a value >0.

cwillisf:
I share @TheLogFather's concerns... specifically:
- How is it possible that we'd run this block without having an active thread? Shouldn't the hat block be running on an active thread, even if it's only active long enough for the hat block to say "nevermind"? This is probably just a lack of understanding on my part, but maybe you can help me learn :)
- Is there really no way to get from a `Block` to its containing `ScratchObj`? I see `app.viewedObj()` being used all over Block.as so I guess it must be true, but it seems like having such a direct path would be a better solution for this change and IMO would also make the code in Block.as a little less fragile.

sclements:
I think I may have misspoken about how this happens. What I did see when tracking this down is that a watcher will request a variable's value before anything is running or selected.

cwillisf:
Ah, that makes sense. I still think it would be better if there were a way to get to the appropriate variable cache without having to go through a thread reference, but since that seems like a much bigger change I'm happy to :+1: this.

TheLogFather:
I think I understand the circumstances under which it actually causes issues:
It's worth looking at https://scratch.mit.edu/projects/20324557/ to see...

It appears that the only time the exception actually arises is on the very first run. If you go into the project and remove the variable from the timer trigger hat block, then replace it, everything works exactly as expected.

Running in Flash Debugger, it's only the very first trigger that causes the exception. If it triggers without the var in place, then it all works as expected after that *with* the var in place.

Rather than just returning a value of zero for this, I wonder if it may be better to simply prevent edge triggers from being checked in the case there is no active thread - i.e. put a check on https://github.com/LLK/scratch-flash/blob/562b9afc6104798cdba06ddebcafc150a5cf5ac8/src/scratch/ScratchRuntime.as#L89 to only run processEdgeTriggeredHats if activeThread is non null...?

sclements:
That may prevent edge triggers like 'loudness > 10' from triggering if nothing is currently running.

Returning zero is our standard 'something's wrong but we need to return something here' value.

apumpleby:
@sclements: Yes, that's exactly the idea - but only on the *first time* it tries to trigger it. After that, once things have been running, it appears to be fine.

Now, there may be a question over *why* it's fine later on (i.e. why is there no exception, hence suggesting that activeThread is not null, even though there is nothing running)...?

TheLogFather:
Wow, wrong browser - shouldn't reply to these things while I'm meant to be in the middle of concentrating on presentations here... :/

Replying again, but with a couple of extra thoughts...

@sclements: Yes, that's exactly the idea - but only on the first time it tries to trigger it. After that, once things have been running, it appears to be fine.

Now, there may be a question over why it's fine later on - i.e. why is there no exception, hence suggesting that activeThread is not null, despite the fact there is nothing running, *and even after stopAll*...? (Does activeThread actually need nullifying explicitly?)

Thinking about it, there is likely still an issue in the case when no scripts at all get run once GF clicked. If there is nothing at all that has been run before it gets to check that edge trigger, then activeThread will probably remain null. But it seems that as soon as *any* script *anywhere* has been run once (even if it is no longer running), then activeThread is not null, so the edge trigger hat works fine with the var in there.

TheLogFather:
Unfortunately, I suspect #912 makes this argument a bit moot... :(

PR #902, opened 2015-08-21
Support importing animated GIFs into a costume

sclements:
Users seem to be trying to import animated GIFs into costumes about 100 times per day and causing errors.


cwillisf:
:+1:

PR #903, opened 2015-08-23
Correct build.properties

DevanWolf:
We forgot to change this file. We had the 3d render directory in ${basedir}/3d_render_src instead of ${basedir}/src/render3d - This changed file will be correct.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -1,13 +1,11 @@
-#
 # To change properties, put overrides into a local.properties file!
-#
 
 # Flex SDK locations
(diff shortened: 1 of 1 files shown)


sclements:
Actually, we don't need the SRC_DIR_3D at all since we build separate SWFs for 3D support and not.

sclements:
Unfortunately, we forgot to cleanup our build files. Not only is 'SRC_DIR_3D' no longer used, but we've moved to using gradle for builds and I think this file is no longer used.

PR #904, opened 2015-08-25
Make stop button call ScratchExtensions.stop()

cwillisf:
This change requires https://github.com/LLK/scratchr2/pull/2757

The `stop()` method is a new method that will allow ScratchExtensions to
decide whether to call the new `_stop()` method or the old `resetAll()`
method on a per-extension basis.


sclements:
:+1:

PR #909, opened 2015-09-01
Set stage color only if supported

cwillisf:
We were setting the `color` property on the stage with the assumption that it would fail silently on versions of Flash which didn't support that feature. Unfortunately on Linux this throws an exception. This change checks for the property before trying to set it.
This fixes LLK/scratch-flash-offline#80


sclements:
Good catch! Using obj['prop'] usually avoids exceptions when the property hasn't been implemented yet but it looks like the Linux version just won't have it. :+1:

PR #911, opened 2015-09-02
Change the way paint actions are committed on undo

### LINKED ISSUE #678: Redo Does Not Work for Bitmap Rectangles, Text and Bitmap Circle
Server - production 434
offline - v434

1. Go to editor
2. Click on paint new sprite
3. Choose text  from bitmap tools
4. Type something then click undo and then redo

Text does not appear

Expected: text reappears

this also impacts circle bitmap tool as well as rectangle

prio-high because this results in data loss.. @cwillisf or @kaschm  let me know if you disagree

cwillisf:
On undo or redo, we now reactivate the tool that currently looks like
it's active. This means that pressing undo with an action in progress
commits that action just like changing a tool would, reducing the number
of code paths which commit an action. The old way was not working if the
undo button was pressed immediately after drawing an ellipse, a
rectangle, or some text.
This fixes #678

### DIFF
--- a/src/svgeditor/ImageEdit.as
+++ b/src/svgeditor/ImageEdit.as
@@ -824,18 +824,7 @@ package svgeditor {
 			}
 		}
 
-		private function clearSelection():void {
-			if (this is BitmapEdit) {
-				var ot:Boolean = currentTool as ObjectTransformer;
-				var tt:Boolean = currentTool is TextTool;
-				if (ot || tt) {
-					shutdown();
-					if (ot) {
-						targetCostume.undoList.pop(); // remove last entry (added by shutdown)
-						targetCostume.undoListIndex--;
-					}
-				}
-			}
+		protected function clearSelection():void {
 		}
 
 		protected final function recordForUndo(imgData:*, rotationCenterX:int, rotationCenterY:int):void {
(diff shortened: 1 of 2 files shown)


src/svgeditor/ImageEdit.as
```diff
@@ -824,18 +824,7 @@ package svgeditor {
 			}
 		}
 
-		private function clearSelection():void {
-			if (this is BitmapEdit) {
```
cwillisf, line 828: Instead of sticking with this type check, I just made the method `protected` and moved the contents into `BitmapEdit`. But then I changed its contents completely. :P

sclements:
Are you sure this does the proper thing for redos?

cwillisf:
It seems to handle redo properly, yeah. Are there specific cases that you're worried about? Note that using the bitmap-select tool clears the redo stack, so there don't seem to be a lot of cases to test with redo. (Maybe bitmap-select shouldn't clear the redo stack until you make a change, but I would treat that as a separate bug.)

sclements:
:+1:  Yeah, I would say that a bitmap selection is a casual action and shouldn't invalidate the redo buffer.

cwillisf:
Bitmap select bug logged as #918.
Thanks for the review!

PR #913, opened 2015-09-06
Fixes #912 and #468, and is a suggestion for an improvement to PR #901.

### LINKED ISSUE #912: Shouldn't be using activeThread for variable getters in edge triggered hats
This is related to PR https://github.com/LLK/scratch-flash/pull/901, as well as, indirectly, to issue https://github.com/LLK/scratch-flash/issues/468.

The use of activeThread in primVarGet (Interpreter.as) is assuming the variable getter must be in the (target of the) currently active thread. That assumption is not necessarily correct for edge-triggered hat blocks - it's quite possible for it to be in a different sprite to the current activeThread.

If the variable in the edge-triggered hat block is local, this can lead to the creation of a new local variable in the other sprite (the one which has the current activeThread), and incorrect evaluation of that variable.

See demo project here: https://scratch.mit.edu/projects/75380478/

TheLogFather:
It adds a new public var in ScratchRuntime.as, currentDoObj, which can be set during allStacksAndOwnersDo. Only processEdgeTriggeredHats uses it here (since that's causing #901 and #468, referencing a stack which is not the currentThread), but it can potentially be used if there are other such cases (for example, a hacked "when I receive" hat block will likely also run into the same bug).

It's probably not the most elegant thing, but it's straightforward, and it works.
Maybe someone has a better idea...?

### DIFF
--- a/src/interpreter/Interpreter.as
+++ b/src/interpreter/Interpreter.as
@@ -677,11 +677,11 @@ public class Interpreter {
 	// a reference to the Variable object is cached in the target object.
 
 	private function primVarGet(b:Block):* {
-		if (activeThread == null) return 0;
+		var target:ScratchObj = app.runtime.currentDoObj ? app.runtime.currentDoObj : activeThread.target;
 
-		var v:Variable = activeThread.target.varCache[b.spec];
+		var v:Variable = target.varCache[b.spec];
 		if (v == null) {
-			v = activeThread.target.varCache[b.spec] = activeThread.target.lookupOrCreateVar(b.spec);
+			v = target.varCache[b.spec] = target.lookupOrCreateVar(b.spec);
 			if (v == null) return 0;
 		}
 		// XXX: Do we need a get() for persistent variables here ?
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -51,6 +51,7 @@ public class ScratchRuntime {
 	public var lastAnswer:String = '';
 	public var cloneCount:int;
 	public var edgeTriggersEnabled:Boolean = false; // initially false, becomes true when project first run
+	public var currentDoObj:ScratchObj = null;
 
 	private var microphone:Microphone;
 	private var timerBase:uint;
@@ -333,7 +334,7 @@ public class ScratchRuntime {
 	private function processEdgeTriggeredHats():void {
 		if (!edgeTriggersEnabled) return;
 		activeHats = [];
-		allStacksAndOwnersDo(startEdgeTriggeredHats);
+		allStacksAndOwnersDo(startEdgeTriggeredHats,true);
 		triggeredHats = activeHats;
 	}
 
@@ -919,7 +920,7 @@ public class ScratchRuntime {
 		return result;
 	}
 
-	public function allStacksAndOwnersDo(f:Function):void {
+	public function allStacksAndOwnersDo(f:Function,setDoObj:Boolean=false):void {
 		// Call the given function on every stack in the project, passing the stack and owning sprite/stage.
 		// This method is used by broadcast, so enumerate sprites/stage from front to back to match Scratch.
 		var stage:ScratchStage = app.stagePane;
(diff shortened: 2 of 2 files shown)


TheLogFather:
I'd think it highly likely that *any* time it gets value of a var during allStacksAndOwnersDo then it should be using the object from that loop for the evaluation.

If that really is the case, then you could drop the setDoObj boolean and the "o" in the child loop, then just always use currentDoObj instead.

src/interpreter/Interpreter.as
```diff
@@ -677,11 +677,11 @@ public class Interpreter {
 	// a reference to the Variable object is cached in the target object.
 
 	private function primVarGet(b:Block):* {
-		if (activeThread == null) return 0;
+		var target:ScratchObj = app.runtime.currentDoObj ? app.runtime.currentDoObj : activeThread.target;
```
cwillisf, line 680: What do you think about:
1) Flipping this conditional: `var target:ScratchObj = activeThread ? activeThread.target : app.runtime.currentDoObj;`
and
2) Always setting `currentDoObj` in `allStacksAndOwnersDo`
?

The reason I ask is that I'm worried about the possibility of getting in here with both `activeThread` and `currentDoObj` unset.

src/scratch/ScratchRuntime.as
```diff
@@ -927,10 +928,13 @@ public class ScratchRuntime {
 		for (var i:int = stage.numChildren - 1; i >= 0; i--) {
 			var o:* = stage.getChildAt(i);
 			if (o is ScratchObj) {
+				if (setDoObj) currentDoObj = ScratchObj(o);
 				for each (stack in ScratchObj(o).scripts) f(stack, o);
 			}
 		}
+		if (setDoObj) currentDoObj = stage;
 		for each (stack in stage.scripts) f(stack, stage);
+		currentDoObj = null;
```
cwillisf, line 937: :thumbsup: for clearing `currentDoObj` at the end here!

TheLogFather:
>What do you think about:
>1) Flipping this conditional: var target:ScratchObj = activeThread ? activeThread.target : app.runtime.currentDoObj;

No - that would actually end up causing the bug again!

The fundamental problem is that activeThread is *the wrong thing* to use.
Hence the title of issue #912: *Shouldn't* be using activeThread...

See the demo project: https://scratch.mit.edu/projects/75380478/ - it creates the variable in the *wrong* sprite because activeThread is totally unrelated to the script for the hat block it's testing.

cwillisf:
Ah, right, sorry. I'm still worried about the possibility of getting into `primVarGet` with both `currentDoObj` and `activeThread` null, though. I can't think of a way that could happen but I'm not 100% positive.

I have this nagging feeling that the "right" fix would be to pipe the target object all the way through such that it becomes an argument of `primVarGet` (and all other primitives) but that seems pretty heavy-handed and your solution seems like a good way to avoid that. I just want to make sure we're not accidentally introducing a crash in some other circumstance :)

TheLogFather:
I'm not totally clear why activeThread can manage to be non-null even when there are no scripts running - I haven't looked into that one.

It does slightly concern me that it's like that because it's not getting cleared when all scripts have finished running. It means it could actually be using an invalid value for activeThread in here. (Though I don't think that can happen with this fix, because of below...)

The key question is to ask what are the possible ways it can get into primVarGet?

There are two I can think of - when it's in the middle of running a script (in which case activeThread will be set), and when it's checking an expression in a hat block (in which case it should have gone through the allStacksAndOwnersDo to find it - at least that's how it finds them at the moment).

The above two cases are covered with this fix (at least for the edge-triggered hats - hacked "when I receive" is not, though it could easily be). But if you do come up with another way, then certainly it will have to be dealt with!

cwillisf:
> The key question is to ask what are the possible ways it can get into primVarGet?

Yep, exactly :)

I've been digging through all the ways to call a primitive that I could find, and I don't see a way to get there that would have both `currentDoObj` and `activeThread` null with your change. I'm going to go ahead and merge this change.

Thanks for talking it through with me, and making the change in the first place! I really appreciate the help in this tricky part of the codebase!

Issue #850, opened 2015-07-14
[3D] Robo Game Behaves differently


2jour:
Server: v438

OS: Windows 7
Browser: Chrome 43

1  https://scratch.mit.edu/projects/2734299/#player 
 
*  Move right and if you go offstage it will reappear to the left of the stage
*  Move all the to the left and ball is not trapped

2 Staging project 1000005164

Do the same steps as above

* You will see ball not reappearing to the left of stage

![stage](https://cloud.githubusercontent.com/assets/6530036/8680339/bf77b0ce-2a2e-11e5-9557-1bc8b731b8f3.png)

* You will see ball stuck on the left side of stage
![stage](https://cloud.githubusercontent.com/assets/6530036/8680405/22e79ebc-2a2f-11e5-9d0d-c81371ff5a74.png)

If i do a ctrl-m it works as in production

sclements:
https://scratch.ly/projects/1000005164/

2jour:
game looks good in the offline editor

cwillisf:
I see strange behavior at the right edge on Firefox / Windows 8.1 as well.

sclements:
Here's an even better project to demonstrate the "deviant" behavior: https://scratch.mit.edu/projects/68808268/

cwillisf:
The edge-of-stage behavior has been fixed with #916
Collision differences have been logged separately as #917

PR #916, opened 2015-09-08
Fill off-stage areas with white for touch test

cwillisf:
When in 2D, doing a `touching color` test while a sprite is partially off the stage tests against solid white. This change implements the same behavior in 3D, where previously the region would be transparent black.

This is a partial fix for #850.


sclements:
:+1:

PR #922, opened 2015-09-11
More realistic piano sounds

as-com:
The piano sounds in Scratch are sub-optimal. The existing sounds cause playback to be massively out of tune for high or low notes. This is my attempt to make it sound better.

Here's a little (not really) project to test it out: 
https://cloud.asun.co/index.php/s/7YMzU8gFGu0zV8I
And here is what that project is "supposed" to sound like: http://imslp.org/wiki/Special:ImagefromIndex/243596


as-com:
Should I do a similar thing to all of the sounds to keep it consistent?

cwillisf:
@as-com I see your commit comment indicating that these WAV files came from the FluidR3 soundfont. What are the licensing terms for that soundfont? We will need to make sure that the terms are compatible with the ways we distribute Scratch.

@ntlrsk do we have someone on the team who knows music well enough to look at this and decide whether it's something we want? I can make a build based on this PR for the sake of this comparison. The replacements certainly sound better to me. We also need to consider whether we would like to take @as-com up on the offer to adjust other instruments as well.

as-com:
@cwillisf FluidR3 is in the public domain. From the (unreadable) readme:
image

However, the Musescore page says it's MIT license. Either way, it seems fine to me.

Jonathan50:
Lol

thisandagain:
I'd like to hold off on this pending some of the work we are doing with @ericrosenbaum for Scratch Jazz and music / audio blocks at large.

PR #924, opened 2015-09-18
Skip empty bitmap element Bitmap-to-Vector convert

### LINKED ISSUE #415: Object created when switching from a blank bitmap image to vector mode
Costume Editor

**How to reproduce:**
1. Create a new sprite
2. Switch from the bitmap editor to the vector editor in the costume editor
3. Draw a selection box around the middle of the vector editor
A tiny object gets selected.

My assumption is that Scratch creates an object when switching from bitmap to vector mode, and that it creates a blank object when the bitmap editor is blank.

cwillisf:
If we notice that all of the pixels of the bitmap are transparent then there's no need to add a bitmap element to the SVG.
This fixes #415

### DIFF
--- a/src/ui/parts/ImagesPart.as
+++ b/src/ui/parts/ImagesPart.as
@@ -335,7 +335,10 @@ public class ImagesPart extends UIPart {
 		useBitmapEditor(false);
 
 		var svg:SVGElement = new SVGElement('svg');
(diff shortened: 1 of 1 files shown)


cwillisf:
@sclements reviewed this in person. See also #925

2jour:
Isn't this 440.4?

PR #929, opened 2015-09-18
Fix the way jsThrowError logs its message

cwillisf:
This was using the wrong call and causing these errors to be logged as `err` with the actual message stored in the `extraData` field.

This fixes LLK/scratch-flash-online#211


sclements:
:+1:

PR #930, opened 2015-09-18
Check length before magic on sprite/costume upload

cwillisf:
This avoids throwing an exception when attempting to read a very short file, such as a zero-byte file.
This fixes LLK/scratch-flash-online#102


sclements:
:+1:

PR #934, opened 2015-09-21
Make two more strings available for translation

cwillisf:
The 'clean up' and 'when Stage clicked' strings are now available for translation.
This fixes LLK/scratchr2#2785

### DIFF
--- a/src/scratch/BlockMenus.as
+++ b/src/scratch/BlockMenus.as
@@ -157,7 +157,7 @@ public class BlockMenus implements DragClient {
 		// translated. This mechanism prevents translating proper names such as sprite,
 		// costume, or variable names.
 		function isGeneric(s:String):Boolean {
-			return ['duplicate', 'delete', 'add comment'].indexOf(s) > -1;
+			return ['duplicate', 'delete', 'add comment', 'clean up'].indexOf(s) > -1;
 		}
 		switch (menuName) {
 		case 'attribute':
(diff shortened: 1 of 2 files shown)


thisandagain:
R+ :+1:

ntlrsk:
@cwillisf @thisandagain  I still don't see "when Stage clicked" available for translation in the Scratch Blocks project (clean up is now available, thanks!)

cwillisf:
@ntlrsk it may be in a different location. In terms of the POT files, the "when Stage clicked" string is in `editor.pot` instead of `blocks.pot`.

ntlrsk:
Yeah, I searched twice (in both Spanish and Danish) and the phrase "when Stage clicked" is not there on translate.scratch.mit.edu...

ntlrsk:
I mean I searched on Scratch Editor, and Scratch Blocks

cwillisf:
@mewtaylor helped me look into this, and we found the string in Scratch Blocks. However, there's something going on that I don't fully understand.

I went to http://translate.scratch.mit.edu/da/blocks/ and downloaded the Danish `blocks.po` file, and in it I found these lines:
```
#~ msgid "when Stage clicked"
#~ msgstr "når Scene er valgt"
```

I'm not sure what the `#~` prefix means, but I assume that these strings won't work with the `#` commenting them out.

ntlrsk:
Ah - the string was in 1.4 but at first when we launched 2.0 we didn't have the block in English - so John must have commented it out. So it does need to be commented out in the code.

PR #935, opened 2015-09-22
Add strings() method to ScriptsPane

cwillisf:
The 'clean up' string was not being included in the editor exports, so I've added it here explicitly. The 'add a comment' string was already included in the editor exports, but I can't figure out why... it seems best to properly include it here as well.
This is hopefully the last SWF change necessary for LLK/scratchr2#2785


cwillisf:
Accidentally created two PRs; see #936

ntlrsk:
from the German translators:
We don't see the new phrases, such as "clean up" in the scratch editor, "save as a copy", "go to my stuff"

PR #936, opened 2015-09-22
Add strings() method to ScriptsPane

cwillisf:
The 'clean up' string was not being included in the editor exports, so
I've added it here explicitly. The 'add a comment' string was already
included in the editor exports, but I can't figure out why... it seems
best to properly include it here as well.


thisandagain:
R+ :+1:

PR #937, opened 2015-09-24
Open internal extension tips with 'ext:' tag

cwillisf:
Internal extensions have special tips content formatted for the built-in tips window. These pages are accessed by passing a specially formatted tag to the tips API in JavaScript -- for example 'ext:LEGO WeDo' for the WeDo extension. This change makes the editor supply that sort of tag when opening tips for an internal extension. The behavior is unchanged for non-internal extensions.

This fixes LLK/scratchr2#2709, assuming the tag -> URL mapping is correct on the JavaScript side of the tips system.

Testing notes:
This change alters the way tips are loaded for internal extensions in general, so please test tips for both WeDo and PicoBoard.

### DIFF
--- a/src/scratch/PaletteBuilder.as
+++ b/src/scratch/PaletteBuilder.as
@@ -372,8 +372,12 @@ public class PaletteBuilder {
 
 	protected function getExtensionMenu(ext:ScratchExtension):Menu {
 		function showAbout():void {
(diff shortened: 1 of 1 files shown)


mewtaylor:
:+1:

cwillisf:
Merged into release/v440 branch

PR #939, opened 2015-10-01
Record Project Video Feature

rthorn:
New feature found in file menu as "Record Project Video".

**Test Cases:**
- Visuals
  - All visual effects (ghost,brightness,mosaic, etc) should work
  - If webcam video is turned on in the project, it should not show up in the resulting video
  - Full editor video should properly scale with resized window (leaves white space on edge)
  - Projects should appear in full editor mode and scale properly
  - Dragged sprites should show in video
- Sound
  - All project sounds should be heard in video with no static
  - Sounds should be properly timed
  - Sounds played with “play note” blocks should be heard
  - Volume changes in code should be reflected in video (for project sound, not mic)
  - If a recording is started in the middle of a playing sound, the sound should start in the correct place in the video
  - Microphone recording should work, volume affected by volume slider in dialog
- All options in opening dialog should work as described, in any combination (except full editor + high quality)
- Should only be able to take video of your own project (online editor only)
- All text should be translated into other languages
- Functionality should be the same in both offline and online editors
- Functionality should be the same in normal layout and small stage mode and presentation mode
- Functionality should be the same for 2D and 3D projects


src/scratch/ScratchRuntime.as
```diff
@@ -89,49 +145,342 @@ public class ScratchRuntime {
...
+	
+	private var pSound:Boolean;
+	private var mSound:Boolean;
+	private var mouse:Boolean;
+	private var cursor:Boolean;
+	public var fullEditor:Boolean;
+	private var framerate:Number;
+	private var width:int;
+	private var height:int;
+	public var ready:int=-1;
```
cwillisf: I think it would make it easier to understand and maintain this code if you could rename and/or add comments to some of these variables to clarify their meaning to someone who isn't familiar with the video code. I think this is especially important since these variables are in a class that isn't just about video recording.

For example, it might make sense to rename `position` to `videoPosition` or something like that. In the case of `ready` (and maybe others?) it could also make sense to turn it into a pseudo-`enum`, similar to `LogLevel`, `MouseEvent`, and so on, to express the meaning of the individual values.

src/scratch/ScratchRuntime.as
```diff
@@ -54,6 +61,7 @@ public class ScratchRuntime {
 
 	private var microphone:Microphone;
 	private var timerBase:uint;
+	private var tool:String;
```
cwillisf: It looks like this variable might be unused

src/scratch/ScratchRuntime.as
```diff
@@ -77,8 +85,56 @@ public class ScratchRuntime {
...
+				count=2;
+			}
+			else if (tR>=1.25) {
+				secs+=tR;
+			}
+			else {
+				app.refreshStagePart();
+			}
+		}
+		if (recording) { // Recording a YouTube video?
```
cwillisf, line 115: Could you explain what this code section is doing? This is the way I'm currently reading it, some of which I likely guessed wrong :)

```
if (the recording has been going for too long) {
    stop recording but grab this last frame
}
else { // usual case during recording
    if (we haven't fed this frame to the encoder yet, and the frame number is divisible by 2 or 3) {
        feed this frame and its sound data to the encoder
        forget this frame and its sound data
    }
}
if (we haven't fed this frame to the encoder yet, and the framerate is 30) {
    feed this frame and its sound data to the encoder
    forget this frame and its sound data
}
```

If we're not recording at 30fps, do we hold onto frames indefinitely if they aren't divisible by 2 or 3? And by "indefinitely" I guess I mean "until starting a new recording", which seems to be when `clearRecording()` is called.

src/scratch/ScratchRuntime.as
```diff
@@ -89,49 +145,342 @@ public class ScratchRuntime {
...
-		if ((frames.length % 100) == 0) {
-			trace('frames: ' + frames.length + ' mem: ' + System.totalMemory);
+		saveSound();
+		var t:Number = getTimer()*.001-secs;
+		while (t>sounds.length/framerate+1/framerate) {
+			saveSound();
+		}
+		if (cursor) var cursor:DisplayObject = Resources.createDO('videoCursor');
+		if (mouse && app.gh.mouseIsDown) var circle:Bitmap = Resources.createBmp('mouseCircle');
+		var f:BitmapData;
```
cwillisf, line 200: It may be possible to reduce garbage collection overhead by reusing the same `BitmapData` across multiple frames as long as the width and height don't change. I'm not sure if that would be worth the extra complexity or not... just a thought.

src/scratch/ScratchRuntime.as
```diff
@@ -89,49 +145,342 @@ public class ScratchRuntime {
...
+				if(Scratch.app.isIn3D) {
+					var scaler:Number = 1;
+					if (!app.editMode) {
+						scaler *= app.presentationScale;
+					}
+					else if (app.stageIsContracted) {
+						scaler*=0.5;
+					}
+					var e:BitmapData = app.stagePane.saveScreenData();
+					if (scaler==1) f.copyPixels(e, new Rectangle(0,0,486,432),new Point(app.stagePane.localToGlobal(new Point(0, 0)).x, app.stagePane.localToGlobal(new Point(0, 0)).y));
```
cwillisf: Why `486,432`? Should those numbers be given named constants, or maybe calculated from other UI elements' locations and sizes?

src/sound/ScratchSoundPlayer.as
```diff
@@ -160,7 +170,9 @@ public class ScratchSoundPlayer {
 		for (i = 0; i < 4096; i++) {
 			var n:Number = interpolatedSample();
 			data.writeFloat(n);
+			dataBytes.writeFloat(n);
 			data.writeFloat(n);
+			dataBytes.writeFloat(n);
 		}
```
cwillisf, line 174: I think it would be more efficient to say `dataBytes.writeBytes(data);` after the loop -- but only if `evt.data` starts empty, which I'm not sure of.

src/ui/RecordingSpecEditor.as
```diff
@@ -0,0 +1,314 @@
...
+	public function RecordingSpecEditor() {
+		addChild(base = new Shape());
+		setWidthHeight(440, 10);
+
+		addChild(description = makeLabel('Capture and download a video of your project to your computer.\nYou can record up to 60 seconds of video.',14));
+		addChild(notSavedLabel = makeLabel('that the video will not be saved on Scratch.',14));
+		addChild(pleaseNoteLabel = makeLabel('Please note',14,true));
+		addChild(moreLabel = makeLabel('More Options', 14));
+		moreLabel.addEventListener(MouseEvent.MOUSE_DOWN, toggleButtons);
+		description.setTextFormat(new TextFormat(null,null,null,null,null,null,null,null,TextFormatAlign.CENTER,null,null,null,5));
```
cwillisf: null,null,null,null,null,null,null,null,lol,null,null,null,null,null,null,null,null

cwillisf:
I made a few "ease of maintenance" suggestions here and there, and I'd like to understand the `if (recording)` section in `ScratchRuntime` better, but overall I'd say it looks pretty good. Looking forward to getting this feature in the hands of Scratchers! :D

src/sound/ScratchSoundPlayer.as
rthorn, line 174: It seems like evt.data always starts out empty, so I'll change that.

src/scratch/ScratchRuntime.as
rthorn, line 200: This would work for the high quality video with a framerate of 30, but for the other framerates the BitmapData is stored in memory for some time before being written to file.

src/scratch/ScratchRuntime.as
rthorn: Fixed this.

src/ui/RecordingSpecEditor.as
rthorn: Haha, fixed this.

src/scratch/ScratchRuntime.as
rthorn, line 115: I've made some comments in the if (recording) section to clarify what's going on - let me know if they're unclear.

src/scratch/ScratchRuntime.as
rthorn: Thanks, removed the variable.

src/scratch/ScratchRuntime.as
rthorn: I changed most of the variable names to be more explicitly related to video recording, and also made ready use declared constants as values.

rthorn:
I've just committed some changes based on your comments. I think I covered everything, but let me know if there's anything else!

MegaApuTurkUltra:
10/10 would star this PR if I could

Jonathan50:
Yaaaaaaaaaay

PullJosh:
Yaaay!

I'm never using processing again! :D

PullJosh:
Haven't looked at the code... hopefully this is one frame per scratch frame, so it isn't influenced by lag?

cwillisf:
Glad to see so much excitement :)
Thanks for the awesome new feature, @rthorn! Sorry it took so long to get it in the build!

@PullJosh there are a few different settings to record at different framerates in order to handle different situations and computer speeds.

liam4:
my life is complete :)

BookOwl:
O.M.G 
HYPE-OVERLOAD!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
:package:

PullJosh:
@cwillisf Awesome! Can't wait to try it out. :D

TheInitializer:
:+1:

Dylan5797:
:+1: * 2

PullJosh:
*braces for one-upping madness*

Thepuzzlegame:
Why is there a 60 second limit on recording? It's not like the videos are being stored on the Scratch servers.

rthorn:
@Thepuzzlegame The videos are stored in your computer's memory while being recorded, so the 60 second limit is to prevent memory crashes.

Jonathan50:
@rthorn oh ok
Is there anyway to disable it? \(if you know it's OK\)

Thepuzzlegame:
@rthorn Ah, I see.

Hamish752:
Wow, sounds awesome!
The 60 second thing seems annoying though...

Dylan5797:
I found
    -		stagePart.refresh();
    +		stagePart.updateRecordingTools();

Shouldn't stagePart.refresh(); stay?

JelloMonster118:
There are no breaks on the hype train.

so is this recording the project itself (going through scripts, running it, and turning it into a movie file)
or recording the project owner's actions through the player?

The second option would allow for "if [key] pressed" or when sprite clicked capabilities, which really shouldn't be in animations unless they're making buttons like 'start animation'

JelloMonster118:
"Should only be able to take video of your own project (online editor only)"
This is probably to stop people from uploading other people's projects to youtube, right? To stop plagiarism? But what's to stop me from just remixing a cool animation and recording it (and then unsharing it after)?

PullJosh:
@JelloMonster118 You bring up a good point.

But please don't next time. You're ruining the hype train. :package:

JelloMonster118:
i guess there /are/ breaks on the hype train.
:package: :train: :train: :train:

cwillisf:
@JelloMonster118, regarding the ability to record someone else's project by remixing it...
We can encourage it, but in the end one must choose for themselves whether to be polite. :)

JelloMonster118:
@cwillisf yeah... trusting 8 year olds not to plagiarize 'videos' isn't going to work. (In fact, full grown adults do it frequently, taking youtube videos, cutting off the names of the original creators and putting it onto facebook (and getting far more views)!)

I can easily see this being used for wrong.

I guess there's no way to moderate this.
like examples:
you could make it so only original projects could be recorded, but then legitimate remixes are excluded. Plus, a plagiarizer can just download-reupload a project and it will look like it was an original project, then they can record it.
You could make it so new scratchers can't record, but just because someone got promoted doesn't necessarily mean they want to follow the rules.
You could make it so only trusted scratchers have them, and you can gain trust by not getting in any trouble for remixing without changes, but then how long must someone work to gain this trust? and if they have the trust they can just quickly record as many remixes as they can make and put them to youtube before they get their trust revoked.
You could make it so remixes have watermarks for the original creator on them, but what if there was a legitimate remix, now stuck with a big ugly watermark for someone else's name?

it's a good idea, but there are just too many problems. :( THAT DOESN'T MEAN YOU SHOULD REMOVE IT THOUGH DON'T LISTEN TO ME PLEASE THIS IS A VERY NICE FEATURE IN THE HANDS OF RESPONSIBLE PEOPLE.
:train: :train: :train: rip hype train

cwillisf:
Honestly, I don't think there's any reason to stop the hype train :)
Scratchers are generally good folks and I trust most to use this feature responsibly and respectfully.
Sure, there will be exceptions, but we can manage that and it doesn't change the fact that I'm sure this will be used for amazing things!

JelloMonster118:
http://i2.kym-cdn.com/photos/images/newsfeed/000/959/055/73a.gif

:train: :train: :train: :package:

Jonathan50:
> 
> :train: :train: :train: :package:

ftfy  
but really, I doubt random people uploading videos of random Scratch projects will get much attention.

PR #940, opened 2015-10-05
Improve logging, especially for translation errors

cwillisf:
Logging events forwarded to the JS console now include extradata.
The `currentLang` is set in Translator.as before checking for translation problems instead of after.
When a warning is logged due to a block argument mismatch, the current language is included in the event.


sclements:
LGTM

PR #945, opened 2015-10-15
Fix block menu focus

cwillisf:


cwillisf:
@sclements said :+1: in an in-person code review

Issue #942, opened 2015-10-13
Condition block disappears when switching language (en -> he)


morantsur:
Found it is this project: https://scratch.mit.edu/projects/79077572/#editor

When I switch to hebrew this part:
![image](https://cloud.githubusercontent.com/assets/15109618/10460032/88df01a2-71a0-11e5-9c61-6b9fec4fd74e.png)

Become this: 
![image](https://cloud.githubusercontent.com/assets/15109618/10460066/a39eb082-71a0-11e5-819c-413b18f28abb.png)
(the "paused=false" is missing, and the project doesn't work)

Switching back to English, it is still missing from the code, but the project is working. 

When then switching to Arabic some more blocks disappered:
![image](https://cloud.githubusercontent.com/assets/15109618/10460128/e838b51c-71a0-11e5-891b-61060fcf7f2d.png)

colbygk:
Pretty sure the reason for this is the translations issues we've been seeing with translations changing the formatting phrases in the strings.

2jour:
I can repro issue

@cwillisf i am assigning to you . Let me know if too busy

PR #946, opened 2015-10-15
Fix `arrayToBlock()` for null substacks in RTL

cwillisf:
When the user selects a new language, the process of updating the script blocks with new strings involves converting them to an array-based representation and back. The way this process was distinguishing between block arguments and substacks had trouble with `null` values, leading to situations where substacks could be eliminated after being interpreted as extraneous block arguments. The new code uses the number of block arguments to avoid the need for such inspection.
This fixes LLK/scratch-flash#942

### DIFF
--- a/src/blocks/BlockIO.as
+++ b/src/blocks/BlockIO.as
@@ -111,8 +111,8 @@ public class BlockIO {
 			b = new Block(label, spec[1], Specs.blockColor(spec[2]), spec[3]);
 		}
 
-		var args:Array = argsForCmd(cmd, b.rightToLeft);
-		var substacks:Array = substacksForCmd(cmd);
+		var args:Array = argsForCmd(cmd, b.args.length, b.rightToLeft);
+		var substacks:Array = substacksForCmd(cmd, b.args.length);
 		var hadSpriteRef:Boolean;
 		for (var i:int = 0; i < args.length; i++) {
 			var a:* = args[i];
@@ -146,15 +146,15 @@ public class BlockIO {
 		return [spec, undefinedBlockType, 0, op]; // no match found
 	}
 
-	private static function argsForCmd(cmd:Array, reverseArgs:Boolean):Array {
+	private static function argsForCmd(cmd:Array, numArgs:uint, reverseArgs:Boolean):Array {
 		// Return an array of zero or more arguments for the given command.
-		// Skip substacks. Arguments may be literal values or reporter blocks (expressions).
+		// Arguments may be literal values or reporter blocks (expressions).
 		var result:Array = [];
-		for (var i:int = 1; i < cmd.length; i++) {
+		for (var i:int = 1; i < 1 + numArgs; i++) {
 			var a:* = cmd[i];
 			if (a is Array) {
-				// block (skip if substack)
-				if (!(a[0] is Array)) result.push(arrayToBlock(a, 'r'));
+				// block
+				result.push(arrayToBlock(a, 'r'));
 			} else {
 				// literal value
 				result.push(a);
@@ -164,14 +164,13 @@ public class BlockIO {
 		return result;
 	}
 
-	private static function substacksForCmd(cmd:Array):Array {
+	private static function substacksForCmd(cmd:Array, numArgs:uint):Array {
 		// Return an array of zero or more substacks for the given command.
 		var result:Array = [];
-		for (var i:int = 1; i < cmd.length; i++) {
+		for (var i:int = 1 + numArgs; i < cmd.length; i++) {
 			var a:* = cmd[i];
 			if (a == null) result.push(null); // null indicates an empty stack
-			// a is substack if (1) it is an array and (2) it's first element is an array (vs. a String)
-			if ((a is Array) && (a[0] is Array)) result.push(arrayToStack(a));
+			else result.push(arrayToStack(a));
 		}
 		return result;
 	}
--- a/src/blocks/BlockIO.as
+++ b/src/blocks/BlockIO.as
@@ -150,7 +150,7 @@ public class BlockIO {
 		// Return an array of zero or more arguments for the given command.
 		// Arguments may be literal values or reporter blocks (expressions).
 		var result:Array = [];
-		for (var i:int = 1; i < 1 + numArgs; i++) {
+		for (var i:int = 1; i <= numArgs; i++) {
 			var a:* = cmd[i];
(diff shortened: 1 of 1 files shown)


src/blocks/BlockIO.as
```diff
@@ -146,15 +146,15 @@ public class BlockIO {
...
 		// Return an array of zero or more arguments for the given command.
-		// Skip substacks. Arguments may be literal values or reporter blocks (expressions).
+		// Arguments may be literal values or reporter blocks (expressions).
 		var result:Array = [];
-		for (var i:int = 1; i < cmd.length; i++) {
+		for (var i:int = 1; i < 1 + numArgs; i++) {
 			var a:* = cmd[i];
 			if (a is Array) {
-				// block (skip if substack)
-				if (!(a[0] is Array)) result.push(arrayToBlock(a, 'r'));
```
cwillisf, line 157: A `null` substack would be picked up as an arg here. By itself that won't cause problems -- the extra arg would just be ignored. However, when using an RTL language the `reverseArgs` flag would be set in this function, causing the arg list to be reversed below (on line 163). That definitely caused problems :)

sclements:
I think I'll have to see this in the debugger to fully grok it. I'll finish the review tomorrow.

src/blocks/BlockIO.as
```diff
@@ -146,15 +146,15 @@ public class BlockIO {
...
 	}
 
-	private static function argsForCmd(cmd:Array, reverseArgs:Boolean):Array {
+	private static function argsForCmd(cmd:Array, numArgs:uint, reverseArgs:Boolean):Array {
 		// Return an array of zero or more arguments for the given command.
-		// Skip substacks. Arguments may be literal values or reporter blocks (expressions).
+		// Arguments may be literal values or reporter blocks (expressions).
 		var result:Array = [];
-		for (var i:int = 1; i < cmd.length; i++) {
+		for (var i:int = 1; i < 1 + numArgs; i++) {
```
sclements: Maybe we could increment numArgs before entering the loop. Or, the argument could be renamed to cmdSize or something similar.

thisandagain:
This is testing ok in Hebrew (RTL) but I'm not sure what else this could be impacting.

thisandagain:
Restarting CI as it appears to have been an issue with Travis

thisandagain:
@bsb20 Could you also give this a review?

cwillisf:
I realize it's not so obvious what's being fixed here and why it's related to translation... here's a bit more information that will hopefully help:

Each `Block` object contains an array of its arguments. The order of these arguments is such that the first element in the array is the one that is left-most on the screen, not necessarily the "first" argument to the `Block` itself. This means that when we switch between LTR and RTL languages we must reverse the order of the arguments. The block itself is aware of whether it was built in LTR or RTL "mode" and we reverse that by rebuilding the block. This is handled in `updateScriptsAfterTranslation()` in `ScratchObj` with the following code:
```actionscript
		for each (var b:Block in scripts) {
			var newStack:Block = BlockIO.arrayToStack(BlockIO.stackToArray(b), isStage);
```

This is basically `fromJSON(toJSON(block))` except that it uses AS3 arrays instead of JSON. The argument order in array form, by the way, is always LTR (see `normalizedArgs()` in `Block`). `stackToArray()` calls `blockToArray()` to do most of the work, and `arrayToStack()` similarly calls `arrayToBlock()`.

Consider this block structure: `if (4 < 5) then {show} else {hide}`. We have one argument, the conditional expression, and two substacks, one for the "then" branch and one for the "else" branch. After `stackToArray()` that should end up as: `['doIfElse', ['<', 4, 5], ['show'], ['hide']]`. I'll use this example below.

The job of the `argsForCmd()` function, called by `arrayToBlock()`, is to look at the array representing a block and return a new array of only the block's arguments, in either LTR or RTL order, not including the blockspec name ('doIfElse') or any of the substacks. In our example block, the expected result for `argsForCmd()` is `[['<', 4, 5]]`: it's an array of arguments, the first element of which is an array representing a `Block`. Similarly, the expected result of `substacksForCmd()` would be `[['show'], ['hide']]`, though the order of these elements does not depend on the active language. Then, these values would be used to create a new block.

So far, so good. So what was the problem?

Note that there's no distinction between arguments and substacks in the array form. That should be OK as long as we put each piece of data in the right spot during `arrayToBlock()`. That's what was sometimes going wrong in the old code.

The old version of `argsForCmd()` would inspect the elements of the array to determine whether they were arguments or substacks. If it found anything other than an array, then that's easy: that must be a literal and it should be added to the arguments list. If the element is an array, though, then look a little closer: only add it to the arguments list if the first element of the array is not an array itself. An array whose first element is an array indicates a substack, which will be handled later by `substacksForCmd()`.

Unfortunately, this logic didn't correctly handle the case where a substack wasn't populated: `null` isn't an array, so it would be taken as an argument by `argsForCmd()` even if it was meant to be taken as a substack.

Here's an expression that wouldn't be handled correctly: `if (4 < 5) then {_}`. The array form of that is `['doIf', ['<', 4, 5], null, null]`. Because of confusion about the null substacks, the old version of `argsForCmd()` would return `[['<', 4, 5], null, null]` in an LTR language, or `[null, null, ['<', 5, 4]]` in an RTL language. When populating a new block from that, the `doIf` block would see that it only needs the first argument and would work fine in an LTR language, which is probably why this bug survived so long, but in an RTL language it would take the `null` and discard the expression. This is how switching between LTR and RTL languages could cause blocks to disappear.

The new version of `argsForCmd()` takes advantage of the fact that we actually create the new block before we process the arguments and substacks, so we know exactly how many arguments it expects. The new code doesn't do any inspection of the arguments at all: it assumes that since `doIf` expects 1 argument the structure of the array must start with `['doIf', arg, substack?...]`. If the array is malformed with extra elements, those extra elements will always be taken as extra substacks and later ignored by the `doIf` block. The `substacksForCmd()` function was also adjusted to use the argument count instead of inspecting elements, to guarantee that the two functions would always be consistent with each other.

Future improvements:
- I think we should always store block arguments in the same order and adjust only the visual presentation for LTR/RTL languages.
- `argsForCmd()` and `substacksForCmd()` could be further simplified by using something like `Array.slice()`, or even `Array.splice()` if we're OK with modifying the array, and then doing an in-place `arrayToBlock()` recursion call.

thisandagain:
@cwillisf +100000 for the wonderful explanation. R+

PR #947, opened 2015-10-28
Make JS log events richer; improve callServer logs

cwillisf:
- Logging events echoed to JS now have their `extraData` sent over
  intact, so that the objects can be inspected in the JS debugger.
- JSON stringification of event `extraData` is protected with a
  try/catch, in case some non-stringifiable data sneaks in. If
  stringification throws the `extraData` will be replaced with a string
  indicating such.
- JSON stringification has been extended to handle `ByteArray` as if it
  were a string. This probably isn't useful except for logging, but it
  does prevent an exception from being thrown.
- The logging for `callServer` has been simplified and uncommented. It
  had originally been commented out in order to avoid overloading our
  event tracking system. That's still a concern, but now it can be
  logged as a warning: it won't generate a tracking event, but it will
  show up in the logging history if a critical error happens afterward.
  It will also show up in the JS console.

Test cases:
- There should be no noticeable change in the editor.
- At least some of the events logged to the JS console should now allow inspection as JS objects. Exact behavior depends on the browser.
- The string part of an event (time stamp, log level, general description) should always be visible even when the `extraData` object is not expanded.

### DIFF
--- a/src/logging/Log.as
+++ b/src/logging/Log.as
@@ -65,7 +65,14 @@ public class Log {
 
 		var extraString:String;
 		function getExtraString():String {
-			return extraString || (extraString = util.JSON.stringify(extraData));
+			if (!extraString) {
+				try {
+					extraString = util.JSON.stringify(extraData);
+				} catch (e:*) {
+					extraString = '<extraData stringify failed>';
+				}
+			}
+			return extraString;
 		}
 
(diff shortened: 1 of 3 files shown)


thisandagain:
Code & testing looks good in online build. R+

Issue #484, opened 2014-09-09
Speech bubble outlines from ask block appears grey instead of blue


DevanWolf:
In Scratch 1.4, I saw that the speech bubble outlines from ask block appear as blue as normal, but in the current version for some reason, they still appear grey instead of the actual blue, which is why it does not match the outline color of the box below.
![Find the grey border...]
(http://wiki.scratch.mit.edu/w/images/Scratch_2.0_Ask_Bubble_Frame.png)

PR #950, opened 2015-11-10
Removing ask prompt when its thread gets stopped, and fixing ask click-through, and ask-bubble outline (fixes #235, #600 and #484)

### LINKED ISSUE #235: Provide a way to remove ask-and-wait
Once there is an ask-and-wait text box shown, the only way (that I know of!) to get rid of it is "stop all". This leads to horrible hacks like http://scratch.mit.edu/projects/21056899/

I've suggested one possibility which may well have fairly minimal impact: making "stop other scripts in sprite" also remove an ask-and-wait box showing in this sprite (after all, that running block is part of a script in this sprite...)  http://scratch.mit.edu/discuss/post/305512/

Another possibility might be to add a new menu item into the "stop" block (e.g. "stop [asking v]"), which would presumably remove all current ask-and-waits.

Other suggestions/possibilities...?

### LINKED ISSUE #600: Click on Ask's blue done button gets sent through to whatever else is on stage (and with wrong scaling & offset)
See http://scratch.mit.edu/discuss/post/736072/ (and note the comment in the project I reference in that post, as well as the project referenced in the OP of that topic).

The click gets sent through to whatever sprite/clone lies at the point of the click's co-ordinates (but with the wrong offset and scaling, so it can appear to be at a completely different place if in fullscreen).

Am I correct in thinking the solution is just to prevent propagation of the mousedown event if the hitTestPoint is satisfied at line 58 of AskPrompter.as...?

TheLogFather:
Remove ask prompter when clone showing it gets deleted, and when "stop other scripts in sprite/stage" stops the thread that it's running in.

Also, it seems sensible to me that Interpreter.stopAllThreads should remove ask prompts if anything else besides ScratchRuntime.stopAll ever uses it (so ScratchRuntime.stopAll no longer needs to explicitly clearAskPrompts itself).

Fixes #235. (Later commits also fix #600 and #484.)

### DIFF
--- a/src/primitives/SensingPrims.as
+++ b/src/primitives/SensingPrims.as
@@ -208,7 +208,7 @@ public class SensingPrims {
 		if (interp.activeThread.firstTime) {
 			var question:String = interp.arg(b, 0);
 			if ((obj is ScratchSprite) && (obj.visible)) {
-				ScratchSprite(obj).showBubble(question, 'talk', true);
+				ScratchSprite(obj).showBubble(question, 'talk', obj, true);
 				app.runtime.showAskPrompt('');
 			} else {
 				app.runtime.showAskPrompt(question);
--- a/src/scratch/AskPrompter.as
+++ b/src/scratch/AskPrompter.as
@@ -55,7 +55,10 @@ public class AskPrompter extends Sprite {
 	public function answer():String { return input.text; }
 
 	private function mouseDown(evt:MouseEvent):void {
-		if (doneButton.hitTestPoint(evt.stageX, evt.stageY)) app.runtime.hideAskPrompt(this);
+		if (doneButton.hitTestPoint(evt.stageX, evt.stageY)) {
+			app.runtime.hideAskPrompt(this);
+			evt.stopImmediatePropagation(); // stop click going through
+		}
 	}
 
 	private function keyDown(evt:KeyboardEvent):void {
--- a/src/interpreter/Interpreter.as
+++ b/src/interpreter/Interpreter.as
@@ -177,6 +177,7 @@ public class Interpreter {
 				if (t.tmpObj is ScratchSoundPlayer) {
 					(t.tmpObj as ScratchSoundPlayer).stopPlaying();
 				}
+				if (askThread == t) app.runtime.clearAskPrompts();
 				t.stop();
 			}
 		}
@@ -206,6 +207,7 @@ public class Interpreter {
 	public function stopAllThreads():void {
 		threads = [];
 		if (activeThread != null) activeThread.stop();
+		app.runtime.clearAskPrompts(); // seems sensible this should happen here
 		clearWarpBlock();
 		app.runtime.clearRunFeedback();
 		doRedraw = true;
--- a/src/primitives/Primitives.as
+++ b/src/primitives/Primitives.as
@@ -199,7 +199,10 @@ public class Primitives {
 	private function primDeleteClone(b:Block):void {
 		var clone:ScratchSprite = interp.targetSprite();
 		if ((clone == null) || (!clone.isClone) || (clone.parent == null)) return;
-		if (clone.bubble && clone.bubble.parent) clone.bubble.parent.removeChild(clone.bubble);
+		if (clone.bubble && clone.bubble.parent) {
+			clone.bubble.parent.removeChild(clone.bubble);
+			clone.bubble = null; // prevent potential exception inside hideAskBubble (via clearAskPrompts in stopThreadsFor below)
+		}
 		clone.parent.removeChild(clone);
 		app.interp.stopThreadsFor(clone);
 		app.runtime.cloneCount--;
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -493,7 +493,16 @@ public class ScratchRuntime {
 		setTimeout(p.grabKeyboardFocus, 100); // workaround for Window keyboard event handling
 	}
 
+	private function hideAskBubble():void {
+		if (interp.askThread && interp.askThread.target) {
+			if (interp.askThread.target!=app.stagePane && interp.askThread.target.bubble) {
+				if (interp.askThread.target.bubble.style=='ask') interp.askThread.target.hideBubble();
+			}
+		}
+	}
+
 	public function hideAskPrompt(p:AskPrompter):void {
+		hideAskBubble();
 		interp.askThread = null;
 		lastAnswer = p.answer();
 		if (p.parent) {
@@ -512,6 +521,7 @@ public class ScratchRuntime {
 	}
 
 	public function clearAskPrompts():void {
+		hideAskBubble();
 		interp.askThread = null;
 		var allPrompts:Array = [];
 		var uiLayer:Sprite = app.stagePane.getUILayer();
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -136,7 +136,7 @@ public class ScratchRuntime {
 
 //----------
 	public function stopAll():void {
-		interp.stopAllThreads();
+		interp.stopAllThreads();  // this does clearAskPrompts now
 		clearRunFeedback();
 		app.stagePane.deleteClones();
 		cloneCount = 0;
@@ -148,7 +148,6 @@ public class ScratchRuntime {
 			s.clearFilters();
 			s.hideBubble();
 		}
-		clearAskPrompts();
 		app.removeLoadProgressBox();
 		motionDetector = null;
 	}
--- a/src/scratch/TalkBubble.as
+++ b/src/scratch/TalkBubble.as
@@ -24,9 +24,9 @@ package scratch {
 public class TalkBubble extends Sprite {
 
 	public var pointsLeft:Boolean;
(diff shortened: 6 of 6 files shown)


TheLogFather:
One further question that comes from this is if it should also remove the speech bubble when it removes the ask input box, in the case that the speech bubble was produced along with the ask block?

At the moment, if you have a visible sprite ask something, the question shows up as a speech bubble. Using MATU's rebroadcast trick removes the ask prompt, but not the speech bubble.

The change within this PR also removes only the ask prompt when using "stop other scripts", with the speech bubble still remaining. (In the case of a clone, that should remove it automatically, since the clone is getting deleted.)

It's easy enough to remove it yourself with a blank "say" block if you need to, but I was wondering about the possibility of having this remove it automatically, if the bubble came from the ask...

Thoughts...?

cwillisf:
Regarding speech bubbles: is your concern about design or implementation?

Design-wise, I think it's a good idea to remove a speech bubble under these circumstances, but I would defer to @ntlrsk. @ntlrsk, what do you think about @TheLogFather's comment above? https://github.com/LLK/scratch-flash/pull/950#issuecomment-155551722

TheLogFather:
I think removing the speech bubble would require some new ScratchSprite variable which is set to the current sprite when the bubble is created along with an ask box. And it gets nullified if the bubble is created through a normal "say" block in that sprite.

Then it'd just be a case of removing the bubble (from the relevant sprite) within clearAskPrompts if that boolean is set.

I'll have a look at doing this in the next couple of days. If you decide it should remove it, I can commit it into here.

cwillisf:
@TheLogFather that all sounds good to me! I don't see a way to avoid adding extra state for this but I do feel that it's worthwhile in this case.

One suggestion: if the state is on the `TalkBubble` (an "I was created by an ask block" Boolean or something like that) then you get some of your state management for free. Don't feel obligated to follow this suggestion if you see a better way, of course :)

TheLogFather:
Yeah, good idea...

I just had a quick look, and TalkBubble already has "private var style:String", which contains "ask" in the relevant case, so that could be used.

The simplest thing is to just make it public so it can be accessed in ScratchRuntime to decide if the bubble needs removing? (Or I guess there could be a new function added into TalkBubble, something like "hideIfStyle(string)", which would provide more flexibility if such might ever be wanted...)

src/primitives/SensingPrims.as
```diff
@@ -208,7 +208,7 @@ public class SensingPrims {
 		if (interp.activeThread.firstTime) {
 			var question:String = interp.arg(b, 0);
 			if ((obj is ScratchSprite) && (obj.visible)) {
-				ScratchSprite(obj).showBubble(question, 'talk', true);
+				ScratchSprite(obj).showBubble(question, 'talk', obj, true);
```
TheLogFather, line 211: This change fixes #484 (speech bubble outline colour)

src/scratch/AskPrompter.as
```diff
@@ -55,7 +55,10 @@ public class AskPrompter extends Sprite {
 	public function answer():String { return input.text; }
 
 	private function mouseDown(evt:MouseEvent):void {
-		if (doneButton.hitTestPoint(evt.stageX, evt.stageY)) app.runtime.hideAskPrompt(this);
+		if (doneButton.hitTestPoint(evt.stageX, evt.stageY)) {
+			app.runtime.hideAskPrompt(this);
+			evt.stopImmediatePropagation(); // stop click going through
+		}
```
TheLogFather, line 61: This change fixes #600 (ask button click-through)

TheLogFather:
@cwillisf, @ntlrsk:
Here's an implementation question for you...

At the moment, when you answer an ask (press return or hit tick button) Scratch will remove a Sprite's bubble (if showing), *no matter where it came from* - i.e. from the ask itself, or from a "say" or "think" that was done in another script while the ask prompt was showing - see line 219 of SensingPrims.as.

This PR was designed to make the ask prompt (alone) go away when a clone currently showing an ask is deleted, and when "stop other scripts" is used in the same sprite/clone as the ask (i.e. to fix #235).

But with the most recent commit it also makes the bubble go away - but *only* if it came from the ask (so a subsequent "say"/"think" bubble will stay there).

Since that commit also fixed the bubble outline for an ask (becomes blue instead of grey, see #484), there's a distinction between a normal say bubble and an ask bubble.

So... is it the desired behaviour to remove a normal "say" bubble (i.e. not-from-the-ask) when ask is answered? Even if it's now a "think" bubble, too? If so, should this PR also remove any type of bubble when removing an ask prompt with "stop other scripts"?

TheLogFather:
Here's (yet) another interesting edge case to ponder...

A hidden sprite will show its ask text as part of the ask prompt, rather than as a speech bubble.

If there is already a say/think bubble for that sprite (but not visible, because the sprite is hidden), then you can send a broadcast just before the ask, and its receiver can then make the sprite visible while the ask prompt, containing the ask text, is still there.

The result is a sprite with both an ask box that contains the ask text, AND a say/think bubble.

Once you answer the ask (press return, or click tick button), it then removes the ask prompt (with the text it contains), but also removes the potentially unrelated speech bubble. (That's due to line 219 of SensingPrims.as.)

Is that the desired behaviour? -Should the say/think bubble be kept if the ask text was in the ask prompt?

An extra point to note: if the sprite is hidden when the ask prompt is removed, then the behaviour is different - the say/think bubble then stays intact (and becomes visible once the sprite is shown) - and it stays intact *even if that bubble came from the ask itself*!

cwillisf:
Wow, thanks for being so thorough, @TheLogFather! It really helped me think about this in much more depth than I had before. I'm going to try to summarize current behavior here and then suggest some desired behavior. Much of this info comes from your posts -- thanks again!

For the purposes of this discussion, keep in mind that there are three kinds of "bubble" -- specifically, "say", "think", and "ask" bubbles, just as @TheLogFather is describing. If I say "bubble" without specifying then I mean any/all of those. In addition, when I say "ask prompt" I mean the box created at the bottom of the stage by the "ask _ and wait" block.

Note that it's currently difficult to tell the difference between "say" and "ask" bubbles due to #484, but @TheLogFather fixes that with one of the changes in this PR.

Current behavior (prior to this PR):

1. Clicking the stop button removes all clones, all "ask" prompts, and all bubbles.
2. Running a "stop {all}" block is equivalent to pressing the "Stop" button above the stage.
3. Each sprite or clone may have at most one bubble at any given time. Attempting to create a second through any method will replace the existing bubble with the newly-created bubble.
4. An "ask" prompt remembers which sprite or clone created it, and creates an "ask" bubble on that sprite or clone.
5. Answering an "ask" prompt removes *any* bubble from the sprite or clone which ran the "ask" block.
6. Running a "stop {this script}" block has no effect on any "ask" prompt or bubbles. Note that "stop {this script}" can't possibly affect an "ask" prompt since a script *waiting* for an "ask" prompt to be answered can't possibly be running a "stop {this script}" block.
7. Running a "stop {other scripts in sprite}" block has no effect on any "ask" prompt or bubbles.

In my opinion, the key here is that point 7 seems inconsistent with point 1.

It seems to me that the most consistent way to make "stop {other scripts in sprite}" remove the "ask" prompt (fixing #235) would be to make "stop {other scripts in sprite}" act like "stop {all}" but filtered to just the current sprite or clone, and it would not stop execution of the current script. That would mean that "stop {other scripts in sprite}" should:
* remove the "ask" prompt if shown by the sprite or clone running the block, and
* remove any bubble of any type attached to the sprite or clone running the block.

Does that sound like it would cover your needs regarding #235 and the like, @TheLogFather?
Do you see any problem with doing that, @ntlrsk?

TheLogFather:
That's a good summary, with a couple of caveats for points four and five...

4) Just to note there is no bubble if the sprite is not showing at the point the ask prompt is started (or if it's on the stage). In that case the text is included in the ask prompt, so there's no immediate issues about removing the bubble for that ask (but see later...)

5) Answering removes any bubble (say/think/ask) but *only if the sprite is visible* at the point the ask is answered. If it's hidden at that point then it doesn't remove any bubble, even if the bubble *did* come from the ask (i.e. if the sprite was visible at the start of the ask, and so showed an ask bubble).

The combination of above means the following:

1) If you hide a sprite during an ask (not common, I'd imagine, but...) then its *ask text* bubble will still be there once you show it again after the ask was answered.
2) If you show a hidden sprite during an ask (again, not common), and make it say/think something, there will be both the ask text in the ask prompt *and* the say/think text in the bubble. In particular, *both* are removed on answering.

OK, so quite how to deal with those (admittedly rather rare) cases, I'll leave up to you to decide... :)

TheLogFather:
I guess the way to make above consistent would be to simply replace line 219 of SensingPrims.as with:

    if ((obj is ScratchSprite) && (ScratchSprite(obj).bubble)) {
        if (ScratchSprite(obj).bubble.style=='ask') ScratchSprite(obj).hideBubble();
    }

So it then *always* removes an existing ask bubble (independent of sprite visibility), but *only* an ask bubble (leaving a say/think bubble untouched).

Let me know if you want that change committed into this PR...

cwillisf (approved):
I think these changes are good. Even though we've discussed further refinements (see conversation above) I'd like to merge this so we don't lose @TheLogFather's improvements when we implement this stuff in 3.0

PR #956, opened 2015-11-20
Fix non-refresh "stop this script" issue #475

### LINKED ISSUE #475: "stop this script" inside non-refresh custom block can cause strange behaviour
Here's a demo project: http://scratch.mit.edu/projects/25912556/

What I think is happening is that the interpreter gets confused about its repeat loops when it reaches "stop this script" while still within the repeat loop. This means the subsequent repeat loop doesn't get executed correctly.

It's something I've noticed a few times before (over a year ago!), but never tried to pin it down. Here's a couple of references in the forums:
http://scratch.mit.edu/discuss/post/148124/
http://scratch.mit.edu/discuss/post/82330/
http://scratch.mit.edu/discuss/topic/50378/

Here's a project where it happened, and I talk about it in the comments:
http://scratch.mit.edu/projects/15648803/

TheLogFather:
[NOTE: description below no longer applies to the way it works in the final commit - see latest code notes and comments instead...]

Instead of trying to set the StackFrame pointer (var sp in Thread.as) to go straight back to the caller (in returnFromProcedure in Thread.as), this nullifies each block in the StackFrame 'chain' up to the caller, and sets the activeThread's block to null.

This means it uses the 'standard' mechanism in stepActiveThread to go back through the StackFrame chain, because it just thinks all of the stacks before the caller have naturally reached their final blocks.

Fixes #475

### DIFF
--- a/src/interpreter/Interpreter.as
+++ b/src/interpreter/Interpreter.as
@@ -665,8 +665,9 @@ public class Interpreter {
 
 	private function primReturn(b:Block):void {
 		// Return from the innermost procedure. If not in a procedure, stop the thread.
-		var didReturn:Boolean = activeThread.returnFromProcedure();
-		if (!didReturn) {
+		if (activeThread.wantReturnFromProcedure()) {
+			activeThread.block = null;  // script will return 'naturally'
+		} else {
 			activeThread.stop();
 			yield = true;
 		}
--- a/src/interpreter/Interpreter.as
+++ b/src/interpreter/Interpreter.as
@@ -272,7 +272,6 @@ public class Interpreter {
 
 			while (activeThread.block == null) { // end of block sequence
 				if (!activeThread.popState()) return; // end of script
-				if (activeThread.block == null) continue;  // can end up null if procedure did "stop this script"
 				if ((activeThread.block == warpBlock) && activeThread.firstTime) { // end of outer warp block
 					clearWarpBlock();
 					activeThread.block = activeThread.block.nextBlock;
@@ -666,9 +665,7 @@ public class Interpreter {
 
 	private function primReturn(b:Block):void {
 		// Return from the innermost procedure. If not in a procedure, stop the thread.
-		if (activeThread.wantReturnFromProcedure()) {
-			activeThread.block = null;  // script will return 'naturally'
-		} else {
+		if (!activeThread.returnFromProcedure()) {
 			activeThread.stop();
 			yield = true;
 		}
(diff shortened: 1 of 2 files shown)


liam4:
+1+1+1+1+1

This would be so useful

BookOwl:
Please do this!!!!!1!1!1!1!1!!!11

TheLogFather:
I think the latest commit is now much closer to the original code's intention.
The only real difference compared to the original is that it no longer does the final popState in returnFromProcedure, but sets the activeThread's block to null. This makes stepActiveThread think the script has ended so that the final popState happens there instead.
(It means there's no need to nullify all the stackFrame blocks, so avoiding all the popStates that were happening in stepActiveThread with the way I had it before.)
There's clearly something going on in stepActiveThread that it needs to take into account when popping the state, and that's not happening when just doing popState in returnFromProcedure.

src/interpreter/Interpreter.as
```diff
@@ -665,8 +665,7 @@ public class Interpreter {
 
 	private function primReturn(b:Block):void {
 		// Return from the innermost procedure. If not in a procedure, stop the thread.
-		var didReturn:Boolean = activeThread.returnFromProcedure();
-		if (!didReturn) {
+		if (!activeThread.returnFromProcedure()) {
```
TheLogFather, line 668: This change isn't necessary for the fix, but just avoids using the local var.

src/interpreter/Thread.as
```diff
@@ -111,8 +111,8 @@ public class Thread {
 	public function returnFromProcedure():Boolean {
 		for (var i:int = sp - 1; i >= 0; i--) {
 			if (stack[i].block.op == Specs.CALL) {
-				sp = i + 1;
-				popState();
+				sp = i + 1;  // 'hack' the stack pointer, but don't do the final popState here
+				block = null;  // make it do the final popState through the usual stepActiveThread mechanism
```
TheLogFather, line 115: Single line fix! :)
(The rest is just my attempt to explain what it means, so could be removed if you want...)

liam4:
You shouldn't use

```
// 'hack
```

You should use

```
// HACK: ...
```

:package:

src/interpreter/Interpreter.as
cwillisf, line 668: I don't think this particular one was me but I sometimes do this type of thing so that I can inspect the value of a variable in a debugger. Thanks for the cleanup :)

cwillisf:
:+1:

PR #967, opened 2015-12-10
Fix #966 by using currentDoObj in Interpreter.targetObj() & targetSprite()

### LINKED ISSUE #912: Shouldn't be using activeThread for variable getters in edge triggered hats

### LINKED ISSUE #966: Should not assume activeThread in Interpreter.targetObj() and Interpreter.targetSprite()
Basically the same as #912 and #468, but with various primitives other than primVarGet.
They use targetObj() & targetSprite(), which, in turn, assumes activeThread.

The blocks for these other primitives can also be included in hat block expressions, so they get the wrong target too.

See demo project: https://scratch.mit.edu/projects/90835029/

Easy one-liner fixes (given existing doCurrentObj in AllStacksAndOwnersDo) - fixed by #967.

TheLogFather:
This fix means currentDoObj gets used for numerous other primitives that can also have their blocks end up in a hat block (where activeThread is invalid) - just as primVarGet uses currentDoObj in #913 to fix #912 & #468.

For a demo of the bug, see https://scratch.mit.edu/projects/90835029/

### DIFF
--- a/src/interpreter/Interpreter.as
+++ b/src/interpreter/Interpreter.as
@@ -93,8 +93,8 @@ public class Interpreter {
 //		checkPrims();
 	}
 
-	public function targetObj():ScratchObj { return ScratchObj(activeThread.target) }
-	public function targetSprite():ScratchSprite { return activeThread.target as ScratchSprite }
+	public function targetObj():ScratchObj { return app.runtime.currentDoObj ? app.runtime.currentDoObj : activeThread.target }
+	public function targetSprite():ScratchSprite { return (app.runtime.currentDoObj ? app.runtime.currentDoObj : activeThread.target) as ScratchSprite }
 
 	/* Threads */
 


cwillisf:
:+1:

PR #969, opened 2015-12-10
Fix #968 - don't assume args for broadcast will be BlockArg.

### LINKED ISSUE #968: Exception when renaming broadcast if any broadcast contains an expression (also when showing senders/receivers)
Caused because get/set broadcastMsg (in Block.as) assume arg is BlockArg.

More easy fixes... :)

TheLogFather:
Prevents exceptions when a broadcast contains an expression.

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -214,8 +214,8 @@ public class Block extends Sprite {
 	}
 
 	public function get broadcastMsg():String {
-		for each (var arg:BlockArg in args) {
-			if (arg.menuName == "broadcast") {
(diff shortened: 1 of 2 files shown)


cwillisf:
:+1: 

I was briefly tempted to suggest using `arg()` to try to evaluate an expression, but that could do almost anything if you consider extensions and I'm not sure it would be very useful in practice.

Maybe in the future we could introduce some sort of "evaluate this expression without side effects or tell me you can't" method, which we could use whenever the UI would like to safely evaluate an expression like this.

Issue #570, opened 2014-12-15
[3D] Colour-space inconsistency between original & Stage3D players for brightness effect


TheLogFather:
Looks like original uses HSL colour-space, so the "brightness" effect is controlling "lightness" in HSL, meaning that increasing the brightness effect always heads towards white.

However, I suspect the Stage3D player is using the HSV/HSB colour-space, meaning the brightness effect is controlling "value"(/"brightness") in HSV(/HSB) colour-space, hence increasing brightness effect does not always head towards white.

See this thread: http://scratch.mit.edu/discuss/topic/73118/

For differences between HSL & HSV/HSB, refer here: http://en.wikipedia.org/wiki/HSL_and_HSV
In particular, note the picture at top-right of the wikipedia page that shows HSL vs HSV/HSB.

jwzimmer:
Closed https://github.com/LLK/scratch-flash/issues/1294 as a dupe of this issue.

PR #972, opened 2015-12-17
Makes Stage3D brightness effect based on HSL colour-space (rather than HSV).

TheLogFather:
First attempt at fixing #570. Needs testing more thoroughly for possible edge-cases...

The way it works is to use the brightness effect value to determine what percentage to boost the current lightness value towards full lightness (so it heads to white), or to reduce it towards zero (so it heads to black).

As an example, if the brightness effect value is 40 then the lightness of a pixel will go 40% of the way to full lightness from where it currently is. If the brightness effect value is -40 then it will drop the current lightness of a pixel by 40%. This means all colours of a costume end up reaching white/black only at brightness effect +/-100% (except those that start off already white/black, obviously), rather than some getting there before others.

This doesn't give exactly the same as the 2D brightness effect, but it's whole a lot better than the current Stage3D behaviour. (Actually, I prefer the look of this to the current 2D brightness effect!)

There's a considerable chunk of AGAL whose purpose I didn't really understand (limiting certain things to 0.11 and 0.09?), and the algorithm in this PR seems to work fine without it, so I've removed all that for now. If it does serve some purpose then an explanation would be useful! :)

### DIFF
--- a/src/render3d/shaders/fragment.agal
+++ b/src/render3d/shaders/fragment.agal
@@ -140,10 +140,11 @@ ft1 = sat(ft1)
 ft2.z = max(ft1.y, ft1.z) //float v = max(dst.g, dst.b);				v		= ft2.z
 ft2.z = max(ft1.x, ft2.z) //v = max(dst.r, v);
 
-//				float span = v - min(r, min(g, b));
-ft2.w = min(ft1.y, ft1.z) //float span =  min(dst.g, dst.b);			span	= ft2.w
-ft2.w = min(ft1.x, ft2.w) //span = min(dst.r, span);
-ft2.w = ft2.z - ft2.w //span = v - span;
+//				float vmin = min(r, min(g, b);
+//				float span = v - vmin;
+ft2.x = min(ft1.y, ft1.z) //float vmin =  min(dst.g, dst.b);			vmin	= ft2.x
+ft2.x = min(ft1.x, ft2.x) //vmin = min(dst.r, vmin);
+ft2.w = ft2.z - ft2.x //float span = v - vmin;                          span	= ft2.w
 
 //				if (span == 0.0) {
 //					h = s = 0.0;
@@ -151,13 +152,8 @@ ft2.w = ft2.z - ft2.w //span = v - span;
 //					if (r == v) h = 60.0 * ((g - b) / span);
 //					else if (g == v) h = 120.0 + (60.0 * ((b - r) / span));
 //					else if (b == v) h = 240.0 + (60.0 * ((r - g) / span));
-//					s = span / v;
 //				}
-ft3.y = ft2.z == 0 //int v_eq_zero = (v == 0);
-ft3.w = 0.001 * ft3.y //tiny = 0.000001 * v_eq_zero;				tiny	= ft3.w
-ft2.z += ft3.w //v = v + tiny;					Avoid division by zero, v != 0
-
-ft3.x = ft2.w == 0 //int span_eq_zero = (span == 0);		span_eq_zero= ft3.x
+ft5.x = ft2.w == 0 //int span_eq_zero = (span == 0);		span_eq_zero= ft5.y
 ft2.y = ft2.w != 0 //int span_not_zero = (span != 0.0); span_not_zero	= ft2.y
 ft3.x = ft1.x == ft2.z //int r_eq_v = (dst.r == v);				r_eq_v	= ft3.x
 ft4.x = ft1.x != ft2.z //int r_not_v = (dst.r != v);				r_not_v	= ft4.x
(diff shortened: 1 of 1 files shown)


griffpatch:
I don't know but the only reason I could think of is to do with limiting
the colour changes from becoming fully transparent which lead to incorrect
collision detection? I know because when I use the set brightness effect in
paper minecraft I get black halos around all my transparent tiles where
scratch has darkened a previously fully transparent area? Would be great
not to do that for me... but it might have implications?

TheLogFather:
@griffpatch:
OK, yes, interesting thought - I've not looked yet at how his change behaves with pixels that already have some alpha, but I *think* the way I've done it here means that anything which is currently transparent should stay transparent.
The alpha component is removed and ignored until the end, when it's then put back in, updated with the ghost effect value, so I *think* that means it shouldn't flip a transparent pixel to non-transparent and vice-versa...? (But doesn't touching get worked out using the original costume anyway? [-after resize, rotate, and effects which change the shape, of course, but without any coloration effects, or, more particularly, without any ghost effect] -I'd think it *should* do...? And I'm pretty sure that the colour-touching tests against the *original* colours of a sprite, not the colours after colour-effect...?)
Anyway, these are the kind of interactions I had in mind when I mentioned "edge-cases" in the commit comment. So, yes, I will have to experiment with some of these things tomorrow and see what turns up...

TheLogFather:
Now that I actually get around to looking at the pixelbender code again, I see where the current Stage3D code has come from:
https://github.com/LLK/scratch-flash/blob/master/src/filters/hsv.pbk#L26-L71
...and the interesting thing is that it, too, was also 'wrong' (i.e. using HSV rather than HSL) - the clue is in the filename, people... ;)

It's only the transform used in the current 2d code that does brightness as something more like lightness in HSL (rather than as 'value' in HSV).

I guess it does raise the question of what behaviour the ST really wants...? Personally, I think it's more intuitive that increasing the 'brightness' will make all parts of a costume head towards white. And using HSV won't do that (since V is already at max for fully saturated red, blue, green, cyan, yellow, magenta, for example - in fact, for every non-grey colour available from the default costume colour palette...)

Also, answering my own question about that chunk of AGAL I skipped, I can now see from the pixelbender code what the intention is:
https://github.com/LLK/scratch-flash/blob/master/src/filters/hsv.pbk#L44-L50

Not sure I'm so keen on it, really - is it necessary? The 2D transform doesn't have anything like it...

thisandagain:
@cwillisf Based on our discussions from last week I think I'd like to land this. Any thoughts?

cwillisf:
@thisandagain The only concern I have is the removal of the code which pretends black is dark gray, and artificially adds saturation to all grays, in order to ensure that the color effect affects those. In this new code, using the color effect block on an all-grayscale costume would (I assume) have no visible effect. (For the record, my new GLSL version of this shader does include special handling for grays.)

All of that said, the results of this special handling are hard to see at all on near-black grays, and are noticeable but not dramatic on lighter grays:
image

If we're OK with removing that special handling then I think this is fine to merge. Do you think we need to run this by @ntlrsk or @mresmres?

thisandagain:
@cwillisf Great. Yes, let's talk about this briefly today during the editor design meeting.

TheLogFather:
There's also the comparison with the 2d transform:
https://github.com/LLK/scratch-flash/blob/master/src/scratch/ScratchObj.as#L255-L257

Haven't checked how that treats black (or white?) with colour effect, but I don't recall noticing that it tweaks them to create a tint...?

cwillisf:
@TheLogFather I agree - I don't think the color transform has any special handling for cases with low value or saturation.

Here's a better example of the results of the special handling. This is from the prototype Scratch 3.0 renderer since it was much easier to tweak the shader in GLSL than assembly. By the way: thanks for this work, @TheLogFather! ;)

This is a pure-gray gradient (0-255) with `set color effect to 50` applied. On the left is the shader without special handling for grays; on the right is the shader with the special handling. Note the band at the top of the right side of the image.
image

TheLogFather:
The 'special handling' does seems to contradict what the colour effect block actually does, which is to simply *shift* the hue. It feels a bit arbitrary to assign grey to correspond to a red hue (which then shifts with the colour effect).

The problem, of course, is that the concept of 'hue', as part of a whole colour-space, is largely beyond the experience of younger Scratchers. It's manageable for the pen colour, since that gives a fixed colour for each value. But the way the colour effect block works as a hue shift makes it much more confusing.

It means that the colour effect is, in practice, rather hit-and-miss, and consequently not used for much other than garish colour-shifting cycles. (I suppose in that case it does then make more sense to arbitrarily assign grey to something, since that adds to such an 'effect'.)

TBH, though, if you really want a more useful and easily-*comprehendible* colour-based effect, I'd suggest some kind of 'tint' effect like in https://scratch.mit.edu/discuss/post/1898432/

thisandagain:
Thanks for the contribution!

PR #990, opened 2016-01-22
    Fixed segmentation undo/redo bugs

bsb20:
-There were some problems with handling click events
    -The as3 mouse events look at where the mouse ends up, which makes
    some problems
    -This is now tracked manually where necessary


cwillisf:
:+1:

PR #991, opened 2016-01-25
Applying mask should trigger save

bsb20:
-Do an "undoable" save when user clicks away
    -Left the option on the commitMask method because
    a redo option may need to apply a mask but should not
    create an additional undo/redo record


bsb20:
Hold off on merging this for a second, this introduces a bug where an extra undo/redo record is made

bsb20:
This should be all set now

cwillisf:
:+1:

PR #996, opened 2016-01-29
Two bug fixes for segmentation tool

bsb20:
-cropRect could possible have a dimension of 0.
    Now the minimum dimension is SEGMENT_STROKE_WIDTH
    -backdrops should have their backgrounds set to opaque
    white rather than transparent black

### DIFF
--- a/src/svgeditor/tools/BitmapBackgroundTool.as
+++ b/src/svgeditor/tools/BitmapBackgroundTool.as
@@ -197,17 +197,21 @@ public class BitmapBackgroundTool extends BitmapPencilTool{
 			if(super.lastPoint == null){
 				dispatchEvent(new Event(BitmapBackgroundTool.UPDATE_REQUIRED));
 			}
-			if(p.x + SEGMENT_STROKE_WIDTH > segmentationState.xMax){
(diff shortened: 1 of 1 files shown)


bsb20:
Please check my math.  The issue was that `cropRect()` sometimes had a `height` or `width` of 0.  To solve this I made the maximum `x` and `y` locations `bitmapLayerData.width - SEGMENT_STROKE_WIDTH` and `bitmapLayerData.width - SEGMENT_STROKE_WIDTH` respectively.  The minimum height and width of the rectangle was then set to `SEGMENT_STROKE_WIDTH`.  This means you can both enforce minimum rectangle dimensions and not worry about your rectangle exceeding the `bitmapLayerData` dimensions.  This was done with composition of `max` and `min` operations, please check that I didn't get this mixed up.

Eric and Kristen also noticed that a backdrop which is segmented should have the removed area set to opaque white `0xFFFFFFFF` (to be consistent with eraser behavior).  I did this by checking `editor.isScene`, which looked right but I'm not 100% sure that means what I think it means.

bsb20:
I changed `lastPoint` so that the `drawRect` in `cropRect` is a non-zero rectangle contained within the bitmap area.  Then I switched to using `rect.intersection` to find the bounding rectangle, which should eliminate any edge cases.

cwillisf:
:+1:

PR #1003, opened 2016-02-08
Magic wand didn't handle transitions well

bsb20:
```
-When leaving bitmapSelect, the work area was not update until after
creating of the new tool
-Magic wand would then use an old copy of the work area
-Added line to update this copy of the work area when this
transition occurs
```

### DIFF
--- a/src/svgeditor/BitmapEdit.as
+++ b/src/svgeditor/BitmapEdit.as
@@ -247,7 +247,9 @@ public class BitmapEdit extends ImageEdit {
 				// User was editing an object and switched tools, bake the object
 				bakeIntoBitmap();
 				saveToCostume();
-				targetCostume.segmentationState.unmarkedBitmap = workArea.getBitmap().bitmapData.clone();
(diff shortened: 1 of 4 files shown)


src/svgeditor/BitmapEdit.as
```diff
@@ -247,6 +247,7 @@ public class BitmapEdit extends ImageEdit {
 				// User was editing an object and switched tools, bake the object
 				bakeIntoBitmap();
 				saveToCostume();
+				targetCostume.segmentationState.unmarkedBitmap = workArea.getBitmap().bitmapData.clone();
```
cwillisf: As a future enhancement, do you think we could skip this most of the time and only do it when activating the magic wand, or something like that?

cwillisf:
:+1:

cwillisf:
These latest changes look good, but it looks like you accidentally pulled in an extra file (and an extra commit?) which modified `MediaLibrary.as`

PR #1004, opened 2016-02-08
Added error handling for sound uploading

bsb20:
```
-Rather than upload an empty sample, show an error msg
```

### DIFF
--- a/src/ui/media/MediaLibrary.as
+++ b/src/ui/media/MediaLibrary.as
@@ -697,7 +697,13 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
 					sound.loadCompressedDataFromByteArray(data, data.length);
 					MP3Loader.extractSamples(origName, sound, sound.length * 44.1, function (out:ScratchSound):void {
 						snd = out;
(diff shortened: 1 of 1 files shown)


src/ui/media/MediaLibrary.as
```diff
@@ -710,7 +716,13 @@ spriteFeaturesFilter.visible = false; // disable features filter for now
...
 						snd = s;
-						startSoundUpload(s, origName, uploadComplete);
+						if(snd && snd.sampleCount > 0) {
+							startSoundUpload(s, origName, uploadComplete);
+						}
+						else{
+							app.removeLoadProgressBox();
+							DialogBox.notify('Error decoding sound', 'Sorry, Scratch was unable to load the sound '+sndName+'.', Scratch.app.stage);
+						}
 					});
```
cwillisf: Would you mind either factoring the callback out into its own named function (could still be local if you want -- doesn't matter to me) or moving the new code into `startSoundUpload`? That way the error check and message would only be in one place.

cwillisf:
:+1: 
Thanks!

PR #1005, opened 2016-02-08
Prevents Select tool from erasing redo stack

bsb20:
```
-Tracks Select tool state to wait until an actual change is made
-Only allows a Select operation to trigger a save when a change is
made
```

### DIFF
--- a/src/svgeditor/BitmapEdit.as
+++ b/src/svgeditor/BitmapEdit.as
@@ -136,10 +136,11 @@ public class BitmapEdit extends ImageEdit {
 	}
 
 	protected override function selectHandler(evt:Event = null):void {
-		if ((currentTool is ObjectTransformer && !(currentTool as ObjectTransformer).getSelection())) {
+		if ((currentTool is ObjectTransformer && !(currentTool as ObjectTransformer).getSelection() && (currentTool as ObjectTransformer).isChanged)) {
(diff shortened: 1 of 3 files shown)


src/svgeditor/BitmapEdit.as
```diff
@@ -136,10 +136,11 @@ public class BitmapEdit extends ImageEdit {
 	}
 
 	protected override function selectHandler(evt:Event = null):void {
```
cwillisf, line 138: For the sake of readability, could I get you to put a line at the top like `var foo:ObjectTransformer = currentTool as ObjectTransformer;` (with a better name than `foo` of course) and then simplify the rest of the function? I know you're not the one that started the fire but it seems like it's continuing to spread ;)

cwillisf:
:+1:, though if you wouldn't mind cleaning up the `is`/`as` stuff it'd be even better :)

bsb20:
I changed this to use a property the way `segmentationTool` is used, and I only did the most basic replacements.

cwillisf:
Oh! Sorry, I see now how my comments might have been misleading. I only meant to suggest `is`/`as` changes within the `selectHandler()` function in `BitmapEdit.as`. I didn't mean to make you do as much as you did!
I do think your changes made `ImageEdit.as` nicer to read, though!

PR #1006, opened 2016-02-09
Fixed SVG grow/shrink problem

bsb20:
-The listener in question was added with useCapture=true
    -When an SVG element is ungrouped, it will be the target of the
    MOUSE_DOWN event
    -Now, check if the element I'm clicking on would make a non-group
    Selection.  If so, add a listener with useCapture=false


cwillisf:
:+1:

PR #1008, opened 2016-02-11
Hiding ColorPicker where appropriate

bsb20:
```
-Also, Improving editor flow in minimally invasive ways
-Namely, certain UI elements should reappear when you click out of
ObjectTransformer mode
-Also, switching colors should deselect items, making it clear this
will not change the current object
-Doing this ^^ required factoring out true color change events from
other draw property changes
```

### DIFF
--- a/src/svgeditor/DrawPropertyUI.as
+++ b/src/svgeditor/DrawPropertyUI.as
@@ -38,6 +38,7 @@ import uiwidgets.*;
 public class DrawPropertyUI extends Sprite {
 
 	public static const ONCHANGE:String = 'onchange';
+	public static const ONCOLORCHANGE:String = 'oncolorchange';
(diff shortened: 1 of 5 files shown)


src/svgeditor/DrawPropertyUI.as
```diff
@@ -243,6 +248,13 @@ public class DrawPropertyUI extends Sprite {
 		if (shapeUI.visible) updateShapeUI();
 	}
 
+	public function onColorChange():void{
+		sendChangeEvent();
+		dispatchEvent(new Event(ONCOLORCHANGE));
+
+
```
cwillisf, line 255: extra newlines here

cwillisf:
:+1:

thisandagain:
This looks great. Thanks @bsb20

PR #1009, opened 2016-02-13
Add the new WeDo 2.0 extension

cwillisf:
Requires LLK/scratchr2#3371 for online support
- Support extensions whose names contain '.'
- Add the WeDo 2.0 extension to the Extension Library
- Update the name and icon for the WeDo 1.0 extension
- Alter `resetPlugin()` to support a bug fix in the offline editor

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -49,7 +49,9 @@ public class ExtensionManager {
 	protected var extensionDict:Object = new Object(); // extension name -> extension record
 	private var justStartedWait:Boolean;
 	private var pollInProgress:Dictionary = new Dictionary(true);
-	static public const wedoExt:String = 'LEGO WeDo';
+	static public const picoBoardExt:String = 'PicoBoard';
+	static public const wedoExt:String = 'LEGO WeDo 1.0';
+	static public const wedo2Ext:String = 'LEGO WeDo 2.0';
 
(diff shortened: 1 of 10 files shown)


src/extensions/ExtensionManager.as
```diff
@@ -49,7 +49,9 @@ public class ExtensionManager {
 	protected var extensionDict:Object = new Object(); // extension name -> extension record
 	private var justStartedWait:Boolean;
 	private var pollInProgress:Dictionary = new Dictionary(true);
-	static public const wedoExt:String = 'LEGO WeDo';
+	static public const picoBoardExt:String = 'PicoBoard';
+	static public const wedoExt:String = 'LEGO WeDo 1.0';
+	static public const wedo2Ext:String = 'LEGO WeDo 2.0';
```
cpseager, line 54: @cwillisf
Won't renaming original WeDo cause opening issues when loading saved files that already have 

"extensionName": "LEGO WeDo"

in their JSON?

cwillisf, line 54: ...yes. Good point. I'll probably revert this line.

cwillisf, line 54: ...and by "probably" I mean "unless I find a better way" -- not "maybe I'll decide to break all those projects"! :flushed:

Thanks for taking a look!

PR #1017, opened 2016-02-17
Removed topbar tools from microworlds

bsb20:


cwillisf:
:+1:
Thanks for taking this one!

PR #1019, opened 2016-02-18
Translate default sprite and message names

cwillisf:
Fixes LLK/scratch-flash-online#245

### DIFF
--- a/src/scratch/BlockMenus.as
+++ b/src/scratch/BlockMenus.as
@@ -780,8 +780,6 @@ public class BlockMenus implements DragClient {
 			else setBlockArg(selection);
 		}
 		var msgNames:Array = app.runtime.collectBroadcasts();
-		if (msgNames.indexOf('message1') <= -1) msgNames.push('message1');
-		msgNames.sort();
 
 		var m:Menu = new Menu(broadcastMenuSelection, 'broadcast');
 		for each (var msg:String in msgNames) m.addItem(msg);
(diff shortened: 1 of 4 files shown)


bsb20:
Looks fine.  I would just want to be sure that no translations would break the name-collision prevention mechanism.  For example in `ScratchStage.as` the `unusedSpriteName` relies on default names being of the form {WORD}{#}.  Are we guaranteed that this won't change when we start translating?

cwillisf:
@bsb20: Yes and no... it's "guaranteed" by convention but not enforced by the code. If someone enters a translation that doesn't end in a number then `unusedSpriteName` will still "work" but the results won't be as expected.

It's possible that a translation might put the digit in a location other than the end of the word. For example, our Basque (`eu`) translation of "Sprite1" is "1Spritea" right now. In that language, `unusedSpriteName` would turn "1Spritea" into "1Spritea2", "1Spritea3", and so on. Nothing will crash and the names will still be unique, but obviously that's not quite what we wanted. (We don't have this in any of our current translation strings.)

Another possibility is that a translation might not have a digit at all -- for example, our Vietnamese (`vi`) translation for "Sprite1" is "Nhân vật" at the moment. I suspect that's just a translation error, but at any rate `unusedSpriteName` should tack a digit onto the end of that and increment as necessary. That may not be the best way to proceed in Vietnamese, but at least the names will be unique.

I think we'll need to replace `withoutTrailingDigits()` with something a little more flexible to get the best behavior across languages, but I think that could be done as a separate change.

bsb20:
agreed.

Issue #1018, opened 2016-02-17
Blocks - Add "random position" to "go to [_]" block


ntlrsk:
As discussed at Scratch editor meeting, add a menu option that tells the sprite to go to a random x-y position on the Stage. 

This will help beginners who want to quickly play with random position and with some beginner tutorials.

![image](https://cloud.githubusercontent.com/assets/4721032/13123668/4939f030-d58b-11e5-8112-410b306be946.png)

liam4:
I'm not sure, as, after all, there's a fairly simple and self descriptive workaround:

```
go to x: pick random -240 to 240 y: pick random -180 to 180
```

That makes it clear that you're moving the sprite to a random X position between -240 to 240 and a random Y position between -180 to 180.

It's just my opinion, though, and your suggestion would be helpful also for quickly mocking up projects.

ntlrsk:
@liam4 Yes, it's true that's a good way to do it already. But, for beginners it's a lot to explain - what x and y each mean, and there's not an easy way to figure out what the range of x and y are (as well as putting the random block in both x and y). So we think this minimal addition through a menu item will be very helpful for beginners.

thisandagain:
Thanks @ntlrsk !
![](https://media.giphy.com/media/ktvFa67wmjDEI/giphy.gif)

TheLogFather:
Will it take into account the dimensions of the costume, and its centering? (i.e. if the costume is 200 pixels across, and it has x-centre-offset of 50 pixels, then the x co-ord can range -from roughly -340-50 to 340-50 –assuming size is 100%, of course...)

Or might you want to do the opposite, and ensure its costume is kept entirely within the bounds of the prject screen? (-which would make it range from roughly -140-50 to +140-50.)

If the centre offset is taken into account for the ranges, then what about when the costume is rotated...? (And the size also needs taking into account for all of this...)

Or will you just take the easy option and make it always just -240 to +240 and -180 to +180? :)

PR #1020, opened 2016-02-18
Added go to random position block

bsb20:
-required creation of new 'location' menu, since the existing menu
    was used in a couple places


bsb20:
incidentally, this whole mechanism seems pretty brittle.

thisandagain:
Resolves GH-1018

cwillisf:
This all looks good to me. It seems like this is an area of the code that could use some cleanup :)

PR #1025, opened 2016-02-23
Reset requester state when args might have changed

cwillisf:
The `collectArgs()` function runs when anything changes in the arguments of a block or any of its descendants, recursively. If this is called on a Requester block which currently has a value, reset / discard that value in case it's no longer valid given the argument change.

This resolves LLK/scratchx#61

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -622,6 +622,10 @@ public class Block extends Sprite {
 	private function collectArgs():void {
 		var i:int;
 		args = [];
+		if (isRequester && requestState == 2) {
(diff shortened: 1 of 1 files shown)


bsb20:
Looks good, I think this will be fine but is definitely symptomatic of the kinds of problems a 'next' extension system could do better.  Also reminds me of the kind of problem solved by [this](http://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared)... basically its just hard know what your arguments are going to do, but if you're aware there's a danger there it might warrant some special machinery.

PR #1027, opened 2016-02-27
Updated Readme

zubiashahid:
Simply removed some blank lines for a demo.


zubiashahid:
this is great demo.

liam4:
Er, why? The blank line is there to clearly show that there's a paragraph break between the two lines. Sure, it renders the same, but why make the raw markdown harder to read (or write)?

thisandagain:
The linebreak is there to show a paragraph break. This change makes the README file harder to read.

PR #1032, opened 2016-03-04
Issues/93

bsb20:
Support for async extension hat blocks, as discussed with @cwillisf .  Extension hat blocks in the offline editor can now properly handle their "synchronous" javascript calls happening asynchronously.  This fixes  [93](https://github.com/LLK/scratch-flash-offline/issues/93).

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -358,11 +358,11 @@ public class ExtensionManager {
 				request(extName, primOrVarName, args, b);
 				return null;
 			}
-			else if(b.requestState == 2 && b.response as Object){
+			else if(b.requestState == 2){
 				var responseObj:Object = b.response as Object;
 				args.push(responseObj);
 				b.requestState = 0;
-				if(responseObj.hasOwnProperty('predicate')){
+				if(responseObj && responseObj.hasOwnProperty('predicate')){
 					app.externalCall('ScratchExtensions.getReporter', function(v:*):void {
 						value = v;
 					}, ext.name, responseObj.predicate, args);
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -49,7 +49,9 @@ public class ExtensionManager {
 	protected var extensionDict:Object = new Object(); // extension name -> extension record
 	private var justStartedWait:Boolean;
 	private var pollInProgress:Dictionary = new Dictionary(true);
-	static public const wedoExt:String = 'LEGO WeDo';
+	static public const picoBoardExt:String = 'PicoBoard';
+	static public const wedoExt:String = 'LEGO WeDo 1.0';
+	static public const wedo2Ext:String = 'LEGO WeDo 2.0';
 
 	public function ExtensionManager(app:Scratch) {
 		this.app = app;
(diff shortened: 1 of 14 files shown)


src/extensions/ExtensionManager.as
```diff
@@ -212,6 +212,10 @@ public class ExtensionManager {
 					// 'r' is reporter, 'R' is requester, and 'rR' is a reporter forced to act as a requester.
 					spec[0] = 'rR';
 				}
+				if(spec[0] == 'h') {
+					// same as above, but force to use async hat blocks
+					spec[0] = 'hH';
+				}
```
cpseager: @bsb20 This is duplicating the existing issue with rR in a saved .sb2 file not reverting to r when opened online. In usability this means the checkbox (e.g. beside tilt or distance in WeDo) is permanently lost to display the watcher on the stage (even when the sb2 file is opened online). 

Basically a saved .sb2 file can end up containing either 'r' or 'rR' and therefore this check should work both ways - i.e. r to rR if offline and rR to r when online. Same for h and hH.

src/extensions/ExtensionManager.as
```diff
@@ -212,6 +212,10 @@ public class ExtensionManager {
 					// 'r' is reporter, 'R' is requester, and 'rR' is a reporter forced to act as a requester.
 					spec[0] = 'rR';
 				}
+				if(spec[0] == 'h') {
```
cwillisf: This line could use an `else`

src/extensions/ExtensionManager.as
cwillisf: @cpseager @bsb20 I think that's an argument to either:
- make sure that `'rR'` gets saved as `'r'` in an SB2, or
- use another way to decide to use this behavior instead (obviating the need for the `'r'` -> `'rR'` swap in the first place).

I don't think I was aware of this problem. I've filed a new issue to track it, since I didn't find one in the editor repositories: https://github.com/LLK/scratch-flash/issues/1051

cpseager: I can't find the issue either, but the attempted fix was previously noted in the (now obsolete) #756

cwillisf: I think I've fixed this in #1053. If we're all happy with #1053 then this PR may need some adjustment as well, but I don't think it'll take too much.

cwillisf:
:+1: overall, pending changes related to #1053
@bsb20, let me know if you'd like to discuss this or my changes in more detail :)

bsb20:
Tested with the offline editor and the prototype async hat blocks in the online editor, looks good.

src/extensions/ExtensionManager.as
```diff
@@ -342,18 +371,53 @@ public class ExtensionManager {
...
+					value = b.response as Boolean;
+				}
+				else{
+					var responseObj:Object = b.response as Object;
+					args.push(responseObj);
+					if(responseObj && responseObj.hasOwnProperty('predicate')){
+						app.externalCall('ScratchExtensions.getReporter', function(v:*):void {
+							value = v;
+						}, ext.name, responseObj.predicate, args);
+					}
```
cwillisf, line 403: If `responseObj` has no `predicate` property here, is that a violation of the async hat block API?
If so, there should probably be an `else` here which logs a warning to the JS console to help out the extension author.
If not, does this need an `else` which sets `value` to... something? `true` maybe?

bsb20, line 403: Good point, fixed in [this commit](https://github.com/LLK/scratch-flash/pull/1032/commits/0af72a727824aa20d373071f0bb5defd66728834)

cwillisf:
:+1:

PR #1033, opened 2016-03-04
Added valid XML check to SVG decoding

bsb20:
-As a catch all, you must be able to make a valid XML object from
    the (potentially) SVG data
    -Adding this to isSVGData seems to engage the preexisting error
    handling functionality

### DIFF
--- a/src/scratch/ScratchCostume.as
+++ b/src/scratch/ScratchCostume.as
@@ -171,7 +171,14 @@ public class ScratchCostume {
 		data.position = 0;
 		var s:String = data.readUTFBytes(10);
 		data.position = oldPosition;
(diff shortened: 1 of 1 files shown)


cwillisf:
@bsb20, do you know why it says that I changed the title and added and removed the "pr - needs review" label? I don't think I did any of that... it looks like that all happened before I was even aware of the PR. I'm only asking because I've seen this on a few things; it makes me wonder if there's some tool or script that's using my account for its actions...

cwillisf:
As for the change itself... LGTM :+1:

bsb20:
Nope, but I have seen the title change thing happen whenever I assign something to you. I guess I assumed you had some script which did this to ensure you got emails about it or something like that...

PR #1037, opened 2016-03-08
Poor type safety in endCurrentTool caused err. 1096

bsb20:
```
-This is not a permanent fix, but solves the problem while touching
at little as possible, and will not throw even if unexpected types
are received.
```

### DIFF
--- a/src/svgeditor/ImageEdit.as
+++ b/src/svgeditor/ImageEdit.as
@@ -30,6 +30,7 @@ import flash.utils.ByteArray;
 import scratch.*;
 
 import svgeditor.objs.*;
+import svgeditor.objs.ISVGEditable;
 import svgeditor.tools.*;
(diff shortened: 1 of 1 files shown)


src/svgeditor/ImageEdit.as
```diff
@@ -767,12 +768,14 @@ public class ImageEdit extends Sprite {
 			setToolMode("select");
 		}
 		// If the tool wasn't canceled and an object was created then select it
-		if (nextObject && nextObject.width > 0 && nextObject.height > 0 && (nextObject is Selection || nextObject.parent)) {
```
cwillisf, line 770: Your new code doesn't include `|| nextObject.parent`... do you know why that was in the old code?

cwillisf:
:+1:

Issue #1031, opened 2016-03-04
"go to %m.spriteOrMouse" is not translated in any languages on v444.


abee2:
A Japanese scratcher reported me about it. It might confuse scratchers who don't speak English.
And I found "go to" has a new option, "random position". Our translation server doesn't have the entry for it.

abee2:
"when Stage clicked", too.

ntlrsk:
Thanks @abee2 for reporting

@bsb20 @cwillisf @mewtaylor- wondering if it's in process to add "random position" to the translation server (I think within "Scratch Editor" since it's a menu item). 

On the other question -  "when Stage clicked" is in Pootle but wondering why it's appearing in the within the "Scratch Editor" project - I would expect it in "Scratch Blocks"?  

Thanks!
![image](https://cloud.githubusercontent.com/assets/4721032/13549226/42dee118-e2cf-11e5-92b8-6765689635fd.png)

cwillisf:
The "go to %m.spriteOrMouse" string has been replaced with "go to %m.location", which should be available in Pootle as of today. Unfortunately this counts as a separate string, but that was necessary for the new feature. In the future I hope to improve the Editor code so that this sort of change doesn't break existing translations.

I added "random position" at the same time.

In addition, "when Stage clicked" has been moved to the "Scratch Blocks" project instead of the "Scratch Editor" project. Hopefully this will help reduce confusion related to that string.

Hope this helps -- sorry this didn't happen before v444 came out!

abee2:
"%m.spriteOrMouse" of "point towards %m.spriteOrMouse" is not translated in any languages on v448.
It is similar to this issue.

PR #1038, opened 2016-03-08
Move 'when Stage clicked' to blocks strings

cwillisf:
The string "when Stage clicked" was showing up in the "UI" strings export (the "Scratch Editor" project in Pootle) instead of the "commands" export (the "Scratch Blocks" project in Pootle). This change fixes that, resolving #1031.

Note that this only affects string export, so it doesn't need to be deployed to take effect. Translators will see the change as soon as the exported files are imported into Pootle.


cwillisf:
@bsb20 and I discussed this verbally

PR #1045, opened 2016-03-14
Fix 1044 extension separator

### LINKED ISSUE #1044: WeDo 2.0 commit has broken other extensions
@cwillisf the commit: 8d42822aeacb6704a1a0582ed598645d368ef8a2
changed the way it handled indexOf('.') and this has caused strange side effects for other extensions which now no longer work correctly e.g. as an example please see attached image where, because the reporter name contained a '.', the extension name, which never contained a '.', has now become corrupted.

capturedeadextension

cwillisf:
Prior to the WeDo 2.0 extension, Scratch assumed that an extension's name would never contain `.` -- Scratch would split an extension op string into an extension name and an op within that extension by looking for the first `.`.

With the WeDo 2.0 extension, I changed the code to split on the last `.` so that "2.0" in the extension name wouldn't cause problems. Unfortunately this broke extensions which use `.` in their blocks.

This change uses U+001F (Unicode "Unit Separator") as the separator instead of `.` since U+001F is very unlikely to be contained in an extension's name or blocks. The separator and unpacking logic have been centralized in `ExtensionManager` as well.

This resolves #1044

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -105,7 +105,7 @@ public class ExtensionManager {
 	// Retrieve the extension name from `prefixedOp`.
 	// If `prefixedOp` doesn't have an extension prefix, return `null`.
 	public static function unpackExtensionName(prefixedOp:String):String {
-		var separatorPosition:int = prefixedOp.lastIndexOf(extensionSeparator);
+		var separatorPosition:int = prefixedOp.indexOf(extensionSeparator);
 		if (separatorPosition < 0) {
 			return null;
 		}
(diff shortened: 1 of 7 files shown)


src/extensions/ExtensionManager.as
```diff
@@ -49,6 +49,7 @@ public class ExtensionManager {
 	protected var extensionDict:Object = new Object(); // extension name -> extension record
 	private var justStartedWait:Boolean;
 	private var pollInProgress:Dictionary = new Dictionary(true);
+	static public const extensionSeparator:String = '\u001F';
```
cwillisf, line 52: See https://en.wikipedia.org/wiki/C0_and_C1_control_codes#Field_separators

bsb20:
:+1:

thisandagain:
Thanks for patching this @cwillisf

Issue #1048, opened 2016-03-17
play note block plays too long after picking note


ntlrsk:
If you pick a note (from the keyboard) on the play note block, it will play the note for too long. This is noticeable if you set instrument to trombone or other non-piano instrument. 

If you click or play the block again then it starts to work fine. 

This affects new tutorial for the international day of jazz (and Clubs and Camps) that we're planning Build a Band. 

![image](https://cloud.githubusercontent.com/assets/4721032/13859852/ed14d814-ec5b-11e5-8b9c-046fcf11d239.png)

PR #1049, opened 2016-03-21
Changed note picker behavior

bsb20:
-Used to play note for 3 seconds
    -Now runs block.demo which is like clicking the block
    -The effect is, the note plays for the specified length
- Fixes [1048](https://github.com/LLK/scratch-flash/issues/1048)

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -904,7 +904,8 @@ public class Block extends Sprite {
 		//make a test duplicate and exec
 		var b:Block = this.duplicate(false);
 		b.nextBlock = null;
-		Scratch.app.runtime.interp.toggleThread(b, Scratch.app.viewedObj(), 1, true);
+		b.visible = false;
+		Scratch.app.runtime.interp.toggleThread(b, Scratch.app.viewedObj(), 1);
 	}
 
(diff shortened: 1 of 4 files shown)


src/blocks/Block.as
```diff
@@ -900,6 +900,13 @@ public class Block extends Sprite {
 		Scratch.app.runtime.interp.toggleThread(topBlock(), Scratch.app.viewedObj(), 1);
 	}
 
+	public function demo():void{
+		//make a test duplicate and exec
+		var b:Block = this.duplicate(false);
+		b.nextBlock = null;
+		Scratch.app.runtime.interp.toggleThread(b, Scratch.app.viewedObj(), 1, true);
```
cwillisf: Have you tried `b.visible = false` here? It seems like that would eliminate the need for the `toggleThread` change and might help in other ways. For example, it might avoid double-rendering the block.

bsb20:
Yep your change looks better.  Just as a point of reference, the isBackground flag for toggleThread is probably gonna happen as a part of [1032](https://github.com/LLK/scratch-flash/pull/1032).

cwillisf:
:+1:

PR #1053, opened 2016-03-23
Eliminate 'rR' block type hack

### LINKED ISSUE #1051: Offline: Projects saved with 'rR' block spec
Because the offline editor's extension interface is asynchronous by nature, the offline editor needs to perform some trickery to simulate a synchronous reporter (`'r'`) block in an extension.

The current version of this trickery involves changing the block type internally to `'rR'`, which tells the editor to treat the reporter as asynchronous but to communicate with the extension in the `'r'`-block style. This code is here:
https://github.com/LLK/scratch-flash/commit/1fe8d3813cf14c33c30b4d3165564dde42a7474a#diff-b8ffd1b1d6e335abaeea8767edd6c819R205

Unfortunately, if a project with such a block is saved from the offline editor, the SB2 ends up including an `'rR'` block. Loading an SB2 with this sort of block into the online editor causes undesirable behavior -- that is, the reporter just doesn't work.

In my opinion, `'rR'` ending up in the SB2 is the bug, and broken reporters in the online editor are a symptom.

cwillisf:
In the offline editor, reporters must be treated as requesters. This had been implemented by converting `'r'` blocks into `'rR'` blocks, but this ended up "leaking" into SB2s saved from the offline editor, which in turn caused problems with the online editor.

Now, the `Block` constructor calls the new `shouldForceAsync()` method on `ExtensionManager` to determine whether or not to set `forceAsync` (renamed from `forceRequester`). The block type/spec is never altered.

In support of this change, the logic for showing the watcher checkbox was updated to base its decision on the constructed block rather than the block spec.

I tried to keep #1032 in mind while making this change; my hope is that this technique will work well there, too.

This resolves #1051

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -31,19 +31,24 @@
 // sequence from a specification string (e.g. "%n + %n") and type (e.g. reporter).
 
 package blocks {
+import assets.Resources;
+
 import extensions.ExtensionManager;
 
 import flash.display.*;
-	import flash.events.*;
-	import flash.filters.GlowFilter;
-	import flash.geom.*;
-	import flash.net.URLLoader;
-	import flash.text.*;
-	import assets.Resources;
-	import translation.Translator;
-	import util.*;
-	import uiwidgets.*;
-	import scratch.*;
+import flash.events.*;
+import flash.filters.GlowFilter;
+import flash.geom.*;
+import flash.net.URLLoader;
+import flash.text.*;
+
+import scratch.*;
+
+import translation.Translator;
+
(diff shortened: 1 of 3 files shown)


cpseager:
Looks much better, but I can't see any fixup support for opening legacy SB2 files that unfortunately already contain a saved 'rR'?

cwillisf:
It doesn't explicitly look for and fix up existing `'rR'` blocks, but the `arrayToStack` process fixes them during load. Actually I don't see how the problem was slipping through there in the first place, but I can say that _it works for me_<sup>TM</sup> now.

PR #1056, opened 2016-03-29
Fix logic typo in isUnofficialExtensionBlock()

cwillisf:
This function was previously returning `true` for non-extension blocks.

### DIFF
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -621,7 +621,7 @@ public class ScratchRuntime {
 
 	private function isUnofficialExtensionBlock(b:Block):Boolean {
 		var extName:String = ExtensionManager.unpackExtensionName(b.op);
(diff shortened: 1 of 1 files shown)


src/scratch/ScratchRuntime.as
```diff
@@ -621,7 +621,7 @@ public class ScratchRuntime {
 
 	private function isUnofficialExtensionBlock(b:Block):Boolean {
 		var extName:String = ExtensionManager.unpackExtensionName(b.op);
-		return !(extName && app.extensionManager.isInternal(extName));
+		return extName && !app.extensionManager.isInternal(extName);
```
cwillisf, line 624: for reference:

``` as3
    public function isInternal(extName:String):Boolean {
        return (extensionDict.hasOwnProperty(extName) && extensionDict[extName].isInternal);
    }
```

bsb20:
:+1:

Issue #875, opened 2015-07-29
Default Chinese Character Font Size Larger


2jour:
enhancemet 1: block font size is small
Almost all Chinese teachers said the size of block is too small !!
Not like English, in Chinese editor interface, it's very hardly recognized.
We can make a comparison:
![image](https://cloud.githubusercontent.com/assets/8316018/7557690/a7ac2436-f7c8-11e4-9da3-41f7aad42f9b.png)![image](https://cloud.githubusercontent.com/assets/8316018/7557715/f0211898-f7c8-11e4-88b6-a499a24ae696.png)
![image](https://cloud.githubusercontent.com/assets/8316018/7557999/ee6bb4b8-f7cd-11e4-9543-56a4519ab552.png)![image](https://cloud.githubusercontent.com/assets/8316018/7558007/fa4d0a8e-f7cd-11e4-86d3-67b915242c29.png)

However, not all people know that press 'shift' & click menu:
![image](https://cloud.githubusercontent.com/assets/8316018/7557712/db4967a4-f7c8-11e4-996f-671508dfcb32.png)

So please enlarge Chinese font size to comfortable.

#757

thisandagain:
Resolved in #1064

PR #1064, opened 2016-04-06
fix chinese font size been too small. (fix  issue #875)

beardad1975:
I find the issue "Default Chinese Character Font Size Larger #875" hasn't been solved. Because the code change doesn't worked .

The reason that it failed is because the files in the locale folder . We have 2 po files for chinese (zh_CN.po zh-cn.po ) and 2 po files for taiwanese (zh_TW.po zh-tw.po) . So I fixed the array to include all these 4 files.

Also the size of the font is still not big enough. So I increase font size a little bit.

I built it in my machine and it worked ! Hope the code could help students in chinese and taiwan for learning Scratch. #875

### DIFF
--- a/src/translation/Translator.as
+++ b/src/translation/Translator.as
@@ -41,7 +41,7 @@ public class Translator {
 	public static var rightToLeft:Boolean;
 	public static var rightToLeftMath:Boolean; // true only for Arabic
 
-	private static const font12:Array = ['fa', 'he','ja','ja_HIRA', 'zh_CN'];
+	private static const font12:Array = ['fa', 'he','ja','ja_HIRA', 'zh_CN', 'zh-cn', 'zh_TW', 'zh-tw'];
 	private static const font13:Array = ['ar'];
 
 	private static var dictionary:Object = {};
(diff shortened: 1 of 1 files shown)


beardad1975:
Anybody knows why the build test failed?

thisandagain:
@beardad1975 Looks like the build failed due to an issue downloading the Flex SDK from Apache (not a problem with your code). I just restarted the build and it passed. @cwillisf would you mind doing a code review?

cwillisf:
Looks good. I tried it out and agree that the font increase is appropriate. The Events category makes it immediately obvious with, for example, the broadcast blocks.
Thanks, @beardad1975 !

@ntlrsk I think we should consider increasing the font size even further for several non-Latin scripts, but I'd like the opinions of native readers.

beardad1975:
Thnak you  so much ! It's help a lot.

2016-04-12 3:08 GMT+08:00 Chris Willis-Ford <notifications@github.com>:

erichuang2013:
nice to see progress here. looking forwarding to see bigger Chinese font.

PR #1067, opened 2016-04-07
Fix `forceAsync` logic

### LINKED ISSUE #1066: Upgrade V445  pick random cant work
After updated v445, My scratch some block cant work.  for example ,  pick random ,mouse X, mouse Y

cwillisf:
A block should only be `forceAsync` if the block is from a JS extension and running in a context where the JS interface is async. Currently the only such context is the offline editor.
Built-in blocks and HTTP extension blocks should never `forceAsync`.

This resolves #1066 and a variety of other issues.


bsb20:
:+1:

Issue #1068, opened 2016-04-07
445 Load of Project with extension blocks is still broken - come up undefined


MrYsLab:
If a new project is created, saved and reloaded, the blocks eventually come up properly. For project that were saved prior to 443, the blocks still come up undefined.

This is causing my users a great deal of angst. Please fix this ASAP.

thisandagain:
Hi @MrYsLab can you please share a SB2 file that demonstrates the problem?

MrYsLab:
Here is a [link](https://drive.google.com/file/d/0B4Qt0LRbWv31d1NseFZ3T0Q4VzQ/view?usp=sharing) to file that demonstrates the problem. The original project was saved with an older version of the editor. I added a single block and resaved using 445.1. As you can see the older information comes up undefined and the new block is defined.

MrYsLab:
Installation of version 445.2 seems to have fixed the problem.

Issue #1063, opened 2016-04-05
v444.4 offline editor, saved project with extension blocks, blocks come up udefined


MrYsLab:
In the past, projects such as s2a_fm containing custom extension blocks could be saved and reopened without issue. Now, after opening a saved project containing extension blocks, all of the extension blocks are red and undefined.

thisandagain:
Tracking in #1068

PR #1071, opened 2016-04-08
Allow both old and new extension separator on load

### LINKED ISSUE #1070: Change to extensionSeparator breaks saved .sb2 programs
Changing the extension separator to \u001f in 8d149c6eb517101b7c64491556deb77bb35200bd (Scratch version 445) breaks saved .sb2 files created with previous versions. It is embarrassing to discover this in front of a class of ten-year-olds, after someone has kindly clicked on the update button. It is not really feasible to ask teachers and students to patch all of their saved programs.

The Scratch version is saved in the .sb2 file in `info.swfVersion` and could be used to infer the extension separator for backwards compatibility. Then the conversion process could simply be, open the old program, then resave it.

We observed this in offline Scratch. Since many of our elementary school classrooms may not use the Internet, and running a local Scratch X server is not within the means of most of our classes, offline Scratch remains a necessary cog in the machine.

On the left, a program in v430; on the right, the same program in v445.
v430 versus v445

cwillisf:
This fixes backward-compatibility for SB2 files saved with extension blocks prior to v445.

This resolves #1070, #1068, and #1063.

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -109,12 +109,13 @@ public class ExtensionManager {
 		for each (var ext:ScratchExtension in extensionDict) {
 			for each (var spec:Array in ext.blockSpecs) {
 				if (spec.length > 2) {
-					var opMatch:Boolean = ext.useScratchPrimitives ?
-							spec[2] == op :
-							(ext.name + extensionSeparator + spec[2] == op) ||
-							(ext.name + extensionSeparatorLegacy + spec[2] == op);
-					if (opMatch) {
-						return [spec[1], spec[0], Specs.extensionsCategory, op, spec.slice(3)];
+					var compareOp:String = ext.useScratchPrimitives ?
+							spec[2] : (ext.name + extensionSeparator + spec[2]);
+					var legacyOp:String = ext.useScratchPrimitives ?
+							spec[2] : (ext.name + extensionSeparatorLegacy + spec[2]);
+					if (op == compareOp || op == legacyOp) {
+						// return using compareOp to upgrade from legacy separator
+						return [spec[1], spec[0], Specs.extensionsCategory, compareOp, spec.slice(3)];
 					}
 				}
 			}
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -50,6 +50,7 @@ public class ExtensionManager {
 	private var justStartedWait:Boolean;
 	private var pollInProgress:Dictionary = new Dictionary(true);
 	static public const extensionSeparator:String = '\u001F';
+	static public const extensionSeparatorLegacy:String = '.';
 	static public const picoBoardExt:String = 'PicoBoard';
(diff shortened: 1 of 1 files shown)


cwillisf:
This will do the wrong thing in some situations, but only I believe those situations are all highly contrived and unlikely. For example, if your project was saved by Scratch >= 445 and it contains an extension called "myExt" with a block called "something.foo" and an extension called "myExt.something" with a block called "foo". In this situation, the legacy rule could cause a problem.

The biggest down-side that I see in this approach is that the project will retain the '.' notation when saved. The code I've changed here can't easily modify the spec of the block being parsed because at this point we're several layers deep in the JSON-parsing code, and the block's actual spec is only mutable in the first layer of that parsing process. IMO that's unfortunate, but not a deal-breaker.

I thought about trying to manipulate the JSON directly, or something else at a higher level than this, but I think that sort of approach would require duplicating a lot of the `arrayToBlock` code or (much more risky) just treat the JSON as a string and assume that substrings of the shape "extension.op" need to be "fixed" without regard to their meaning. IMO, the `specForCmd` approach is the lesser evil.

cwillisf:
Note: the Travis build is failing because the Apache Flex SDK web site is down for maintenance.

cwillisf:
I figured out a way to update the separators on load. As of 4ca408002e94b53c76cc5250468eff39411bc481 a load/save cycle will result in a saved SB2 which uses U+001F as the separator.

src/extensions/ExtensionManager.as
```diff
@@ -106,10 +107,16 @@ public class ExtensionManager {
...
-			var prefix:String = ext.useScratchPrimitives ? '' : (ext.name + extensionSeparator);
 			for each (var spec:Array in ext.blockSpecs) {
-				if ((spec.length > 2) && ((prefix + spec[2]) == op)) {
-					return [spec[1], spec[0], Specs.extensionsCategory, op, spec.slice(3)];
+				if (spec.length > 2) {
+					var compareOp:String = ext.useScratchPrimitives ?
+							spec[2] : (ext.name + extensionSeparator + spec[2]);
+					var legacyOp:String = ext.useScratchPrimitives ?
+							spec[2] : (ext.name + extensionSeparatorLegacy + spec[2]);
+					if (op == compareOp || op == legacyOp) {
```
cwillisf, line 116: The comparison against `legacyOp` is redundant but harmless if `ext.useScratchPrimitives` is true. It's important that `legacyOp == compareOp` in that case, though, to make sure that the comparisons are identical.

Suppose, for example, that `legacyOp` was initialized to `null` or `undefined` in the `ext.useScratchPrimitives` case. This function could then return an unexpected spec -- specifically, whatever happens to be the first block spec in the first extension -- when called with a `null` or `undefined` op. This function should never be called with a `null` or `undefined` op, but if we ever have such a bug I don't want to make it harder to track down.

thisandagain:
This is great. Thanks for the update on the original fix. This LGTM. :+1:

PR #1084, opened 2016-04-20
Fix RTL block menus

### LINKED ISSUE #984: Placeholders in RTL languages revert when changed and break the block
Reported by a user for hebrew: https://mitscratch.freshdesk.com/helpdesk/tickets/37151

If you use the sensing 'x position' of 'sprite 1' block, then switch into hebrew it looks correct:
'sprite 1'  של 'מיקום על ציר x'
But if you then select the sprite 1 pull down to select a different sprite (or even the same one) the text reverts to 'x-position'.
So you end up with something that says
'x-position in english' of  'x-position in hebrew'
And the block no longer works.

Simple project to demonstrate: https://scratch.mit.edu/projects/91831001/#editor

cwillisf:
The code for the `getAttribute:of:` block was not RTL-aware and had been mixing up the two menus on the block. Fixing this led me to discover a few other RTL-related issues as well.

This resolves #984 as well as the following:
- Changing from an LTR language (like English) to an RTL language (like Hebrew) with the `{X position} of {Stage}` block in the scripting area would change "Stage" to "_stage_" instead of its proper translation. This could happen for some other blocks and other targets as well. Here's the list of blocks and targets which run through that code, though only some of these were affected by the problem:
  - blocks: `createCloneOf`, `distanceTo:`, `getAttribute:of:`, `gotoSpriteOrMouse:`, `pointTowards:`, `touching:`
  - targets: `_edge_`, `_mouse_`, `_myself_`, `_stage_`, `_random_`
- In drop-down menus which allow the user to select the stage (such as `{X position} of {Stage}`), the word "Stage" was always shown in English.

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -1609,7 +1609,15 @@ public class Scratch extends Sprite {
 
 	public function externalCall(functionName:String, returnValueCallback:Function = null, ...args):void {
 		args.unshift(functionName);
-		var retVal:* = ExternalInterface.call.apply(ExternalInterface, args);
+		var retVal:*;
+		try {
+			retVal = ExternalInterface.call.apply(ExternalInterface, args);
+		}
+		catch (e:Error)
+		{
+			logException(e);
+			// fall through to below
+		}
 		if (returnValueCallback != null) {
 			returnValueCallback(retVal);
 		}
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -240,6 +240,19 @@ public class Block extends Sprite {
 		}
 	}
 
+	// Convert a left-to-right argument index into a current argument index:
+	// - If the block is LTR, then return the index as it is,
+	// - Otherwise count back from the end to return the new index.
+	public function getNormalizedArgIndex(ltrIndex:int):int {
+		return rightToLeft ? args.length - 1 - ltrIndex : ltrIndex;
+	}
+
+	// Retrieve the argument which would have the given index in LTR mode,
+	// regardless of whether this block is currently LTR or RTL.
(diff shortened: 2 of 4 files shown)


src/Scratch.as
```diff
@@ -1609,7 +1609,15 @@ public class Scratch extends Sprite {
...
-		var retVal:* = ExternalInterface.call.apply(ExternalInterface, args);
+		var retVal:*;
+		try {
+			retVal = ExternalInterface.call.apply(ExternalInterface, args);
+		}
+		catch (e:Error)
+		{
+			logException(e);
+			// fall through to below
+		}
```
cwillisf, line 1620: This makes it easier to debug localization issues when not connected to a server.

src/blocks/Block.as
```diff
@@ -240,6 +240,19 @@ public class Block extends Sprite {
...
+	public function getNormalizedArgIndex(ltrIndex:int):int {
+		return rightToLeft ? args.length - 1 - ltrIndex : ltrIndex;
+	}
+
+	// Retrieve the argument which would have the given index in LTR mode,
+	// regardless of whether this block is currently LTR or RTL.
+	public function getNormalizedArg(ltrIndex:int):* {
+		return args[getNormalizedArgIndex(ltrIndex)];
+	}
+
```
cwillisf, line 255: I could use `normalizedArgs()` (below) but I'd rather not reverse the whole args array when I only want one element.

thisandagain:
This is great. Thanks @cwillisf 

@jwzimmer this is something we should really focus on in the test plan, but specific to this PR I'd also like to make sure we test L10N in the editor prior to release to ensure we don't have regressions.

PR #1102, opened 2016-05-03
Handle image load errors more gracefully

### LINKED ISSUE #1085: Projects stuck on 100% of assets loaded; can't "see inside" them
Two users have sent issues to freshdesk saying they cannot load their projects. The projects show the assets being loaded, and eventually all are loading according to the progress bar/ UI, but the project seems to hang there. You cannot load it in the editor via My Stuff or "See Inside" button. The project does not play.

User 1:
- Project ID: 105311331
- Operating System: MacOS Macintosh X 10.10.5 
- Browser: Chrome 49.0.2623.112 
- Flash Version: Flash 21.0 (release 0)

User 2:
- Project ID: 104012599 
- Operating System: Windows NT 10.0 
- Browser: Chrome 49.0.2623.112 
- Flash Version: Flash 21.0 (release 0)

I could not load the projects, either. To reproduce, just go to the project ID using the Scratch URL.
screen shot 2016-04-21 at 11 39 43 am

scratch v445.2
scratch-flash 2a019c6
scratch-flash-online 015f4cc
Browser: Chrome 50.0.2661.75 (64-bit)

cwillisf:
Prior to this change, any error encountered on any image (costume or backdrop) in the project would abort the `decodeAllImages` process. Unfortunately this was handled in such a way that it would cause the project to never finish loading, resulting in projects getting stuck at 100% on the loading progress bar.

Now, images which fail to load are replaced with an empty bitmap costume. In addition, such failures are reported to the user in a dialog at the end of the loading process. This report includes the sprite name (if applicable) and the costume/backdrop name for each failure.

This may also cause a small load-time speedup for projects with many costumes or backdrops due to the elimination of the `for each` in `decodeAllImages/imageDecoded()`.

This resolves #1085.

Things to test:
- The projects mentioned in #1085 should now finish loading and mention which costume was corrupted.
- Loading projects from "My Stuff" should work correctly: bitmap and vector costumes should appear as expected.
- A sprite or costume imported from a bitmap (PNG, JPG, etc.) or SVG should appear in the editor as expected.
- After any of the above tests, it should be possible to load a different SB2 file. If nothing happens when trying to load the SB2 file, that implies the previous project load process never finished.

@bsb20 @sclements
The more eyes on this, the merrier :)

### DIFF
--- a/src/svgutils/SVGImporter.as
+++ b/src/svgutils/SVGImporter.as
@@ -271,7 +271,7 @@ public class SVGImporter {
 			imagesLoaded++;
 			if ((imagesLoaded == imageCount) && (whenDone != null)) whenDone(root);
 		}
-		var imageCount:int, imagesLoaded:int;
+		var imageCount:int = 0, imagesLoaded:int = 0;
 		for each (var el:SVGElement in root.allElements()) {
 			if (('image' == el.tag) && (el.bitmap == null)) {
 				imageCount++;
--- a/src/util/ProjectIO.as
+++ b/src/util/ProjectIO.as
@@ -55,8 +55,22 @@ public class ProjectIO {
 		this.app = app;
 	}
 
+	private static var translationStrings:Object = {
+		imageLoadingErrorTitle: 'Image Loading Error',
+		imageLoadingErrorHeader: 'At least one backdrop or costume failed to load:',
+		imageLoadingErrorBackdrop: 'Backdrop: {costumeName}',
+		imageLoadingErrorSprite: 'Sprite: {spriteName}',
+		imageLoadingErrorCostume: 'Costume: {costumeName}'
+	};
+
 	public static function strings():Array {
-		return [];
+		var result:Array = [];
+		for each (var key:String in translationStrings) {
+			if (translationStrings.hasOwnProperty(key)) {
+				result.push(translationStrings[key]);
+			}
+		}
+		return result;
 	}
 
 	//----------------------------
@@ -243,20 +257,28 @@ public class ProjectIO {
 		var errorCostumes:Vector.<ScratchCostume> = new Vector.<ScratchCostume>(2);
 
 		function makeErrorImage(obj:ScratchObj, c:ScratchCostume):* {
-			// TODO: translation?
 			if (!errorDialog) {
 				errorDialog = new DialogBox();
-				errorDialog.addTitle('Image loading error');
-				errorDialog.addText('At least one backdrop or costume failed to load:\n');
-			}
-			if (obj.isStage) {
-				errorDialog.addText('Backdrop: ' + c.costumeName + '\n');
-			}
-			else {
-				errorDialog.addText('Sprite: ' + obj.objName + '\nCostume: ' + c.costumeName + '\n');
+				errorDialog.addTitle(translationStrings.imageLoadingErrorTitle);
+				errorDialog.addText(Translator.map(translationStrings.imageLoadingErrorHeader) +'\n');
 			}
 
-			var errorCostumeIndex:int = 0 + obj.isStage;
+			var itemText:String;
+			if (obj.isStage) {
+				itemText = Translator.map(translationStrings.imageLoadingErrorBackdrop);
+			}
+			else {
+				itemText = Translator.map(translationStrings.imageLoadingErrorSprite) + '\n' +
+						Translator.map(translationStrings.imageLoadingErrorCostume);
+			}
+
+			var context:Dictionary = new Dictionary();
+			context['spriteName'] = obj.objName;
+			context['costumeName'] = c.costumeName;
+			itemText = StringUtils.substitute(itemText, context);
+			errorDialog.addText(itemText + '\n');
+
+			var errorCostumeIndex:int = int(obj.isStage);
 			var errorCostume:ScratchCostume = errorCostumes[errorCostumeIndex] =
 					errorCostumes[errorCostumeIndex] || ScratchCostume.emptyBitmapCostume('', obj.isStage);
 			return errorCostume.baseLayerBitmap;
--- a/src/util/ProjectIO.as
+++ b/src/util/ProjectIO.as
@@ -207,48 +207,113 @@ public class ProjectIO {
 		}
 	}
 
+	// Load all images in all costumes from their image data, then call whenDone.
 	public function decodeAllImages(objList:Array, whenDone:Function, fail:Function = null):void {
-		// Load all images in all costumes from their image data, then call whenDone.
-		function imageDecoded():void {
-			for each (var o:* in imageDict) {
-				if (o == 'loading...') return; // not yet finished loading
+		// This should be called on success or failure of each image
+		function imageDone():void {
+			if (--numImagesToDecode == 0) {
+				allImagesLoaded(objList, imageDict, whenDone, fail);
 			}
-			allImagesLoaded();
-		}
-		var error:Boolean = false;
-		function decodeError():void {
-			if (error) return;
-			error = true;
-			if (fail != null) fail();
-		}
-		function allImagesLoaded():void {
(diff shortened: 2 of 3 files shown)


src/svgutils/SVGImporter.as
```diff
@@ -271,7 +271,7 @@ public class SVGImporter {
 			imagesLoaded++;
 			if ((imagesLoaded == imageCount) && (whenDone != null)) whenDone(root);
 		}
-		var imageCount:int, imagesLoaded:int;
+		var imageCount:int = 0, imagesLoaded:int = 0;
```
cwillisf, line 274: This was causing a warning in IDEA, which I encountered when I was verifying that all paths would eventually call `whenDone`.

src/util/ProjectIO.as
```diff
@@ -207,48 +207,113 @@ public class ProjectIO {
...
+				errorDialog.addText('Backdrop: ' + c.costumeName + '\n');
+			}
+			else {
+				errorDialog.addText('Sprite: ' + obj.objName + '\nCostume: ' + c.costumeName + '\n');
+			}
+
+			var errorCostumeIndex:int = 0 + obj.isStage;
+			var errorCostume:ScratchCostume = errorCostumes[errorCostumeIndex] =
+					errorCostumes[errorCostumeIndex] || ScratchCostume.emptyBitmapCostume('', obj.isStage);
+			return errorCostume.baseLayerBitmap;
```
cwillisf, line 284: You might argue that we should replace a broken vector costume with a new, empty vector costume. I agree in principle, but at this point in the code we've already cleared out any information we could use to reliably determine whether or not the costume is supposed to be vector.
Note that `ScratchCostume.isSVGData()` only returns `true` if the data is complete and well-formed SVG, so corrupted SVG data looks like bitmap data if you use that method.

src/util/ProjectIO.as
```diff
@@ -207,48 +207,113 @@ public class ProjectIO {
...
+
+		if (errorDialog) {
+			errorDialog.addButton('OK', errorDialog.accept);
+			errorDialog.showOnStage(Scratch.app.stage);
+
+			if (fail != null) {
+				fail();
+			}
+			else if (whenDone != null) {
+				whenDone();
```
cwillisf, line 331: Some callers provide a failure callback, and some expect `whenDone()` to be called even on failure. Failing to call `whenDone()` in the failure case was basically the core cause of #1085.
While it would be nice to make callers consistent, I chose to do this solution instead of making this PR even larger...

src/util/ProjectIO.as
```diff
@@ -276,13 +340,16 @@ public class ProjectIO {
 			imageDict[svgData] = svgRoot;
 			doneFunction();
 		}
-		if (imageDict[svgData] != null) return; // already loading or loaded
+		if (imageDict[svgData] != null) {
+			// already loading or loaded
+			doneFunction();
+			return;
```
cwillisf, line 368: We _need_ `doneFunction()` to be called on every path, otherwise `decodeAllImages` never calls `allImagesLoaded` and we end up in project-load limbo.

src/util/ProjectIO.as
```diff
@@ -207,48 +207,113 @@ public class ProjectIO {
...
 
-		var c:ScratchCostume;
-		var allCostumes:Array = [];
-		for each (var o:ScratchObj in objList) {
-			for each (c in o.costumes) allCostumes.push(c);
+	private function allImagesLoaded(objList:Array, imageDict:Dictionary, whenDone:Function, fail:Function):void {
+		var errorCostumes:Vector.<ScratchCostume> = new Vector.<ScratchCostume>(2);
+
+		function makeErrorImage(obj:ScratchObj, c:ScratchCostume):* {
+			// TODO: translation?
```
thisandagain: Do we have a process for translating error dialogs?

src/util/ProjectIO.as
thisandagain, line 331: 👍

thisandagain:
This looks good. Do you have a few projects that we can use for testing? If so, it would be nice to attach them to the PR for @jwzimmer.

src/util/ProjectIO.as
```diff
@@ -207,48 +207,113 @@ public class ProjectIO {
...
 
+	// Load all images in all costumes from their image data, then call whenDone.
 	public function decodeAllImages(objList:Array, whenDone:Function, fail:Function = null):void {
-		// Load all images in all costumes from their image data, then call whenDone.
-		function imageDecoded():void {
-			for each (var o:* in imageDict) {
-				if (o == 'loading...') return; // not yet finished loading
+		// This should be called on success or failure of each image
+		function imageDone():void {
+			if (--numImagesToDecode == 0) {
```
thisandagain, line 228: Sorry. I'm not sure I totally grok the design patterns in our AS3 code. Totally ok to keep if this is normal. 
Nit: this feels a tad bit hard to scan for me without a performance benefit AFAIK.

cwillisf, line 228: Ah... just to be clear, I made this change because I wanted to stop storing the magic string 'loading...' in the dictionary which usually contains image data (`BitmapData` or `SVGElement`), and using that to check whether each image is done. The alleged performance benefit was a side-effect :)

The alleged performance benefit: the old code would check every image in `imageDict` for the string `'loading...'` each time any image in `imageDict` finished for a total of O(n^2) string checks. The new code just decrements the integer and checks that against zero.

This form of count-based "is everybody done?" pattern is my usual, but only because I haven't encountered any method that I like better. I'm certainly open to suggestions. For example, some of the SVG code keeps two separate counts, one for submitted requests and one for finished requests, and declares the overall batch to be done when the two counts are equal.

src/util/ProjectIO.as
cwillisf: Not really. Usually the "process" for translating anything in the editor is for me to add it to the `strings()` method, do the editor string export, and make it show up in Pootle.
The main reason I bailed on this was the degree of dynamic text, but that's because I forgot that I added [a way to add placeholders](https://github.com/LLK/scratch-flash/blob/3ed8b522be528b056a1b577630118bb42a95c227/src/uiwidgets/DialogBox.as#L108) to a translatable string. I think we only use that in one place so far (the "Project not saved" dialog in the online editor) but we could certainly use it here as well. I'll give that a shot tomorrow.

src/util/ProjectIO.as
thisandagain, line 228: Thanks for clarifying!

cwillisf:
Here are some SB2 files that can be used for testing. I corrupted the images myself by truncating the files, which is the same sort of corruption I saw in the projects listed on #1085.

I've renamed the SB2 files to .ZIP so that GitHub would accept them as attachments. SB2s are really ZIP files, anyway :)

[Project with corrupt bitmap backdrop.zip](https://github.com/LLK/scratch-flash/files/249344/Project.with.corrupt.bitmap.backdrop.zip)
[Project with corrupt bitmap costume.zip](https://github.com/LLK/scratch-flash/files/249345/Project.with.corrupt.bitmap.costume.zip)
[Project with corrupt vector backdrop.zip](https://github.com/LLK/scratch-flash/files/249346/Project.with.corrupt.vector.backdrop.zip)
[Project with corrupt vector costume.zip](https://github.com/LLK/scratch-flash/files/249347/Project.with.corrupt.vector.costume.zip)

thisandagain:
This is perfect. Thanks @cwillisf !

rschamp:
I downloaded project 106997819 with @cwillisf's nifty script, and uploaded it to scratch.ly. I can play the project, and it informed me that the backdrop failed to load.  The failed backdrop appears in the editor as blank.  Everything else in the project works as expected.

[106997819.sb2.zip](https://github.com/LLK/scratch-flash/files/287022/106997819.sb2.zip)

PR #1105, opened 2016-05-06
Check more carefully for Scratch stage clicks

### LINKED ISSUE #1100: Double-clicking in Sound Library clicks through to the Stage
Double-clicking in the Sound Library clicks through to the Stage - this can switch which sprite is selected -- which is problematic.  (We saw this happen with kids in the music workshop on Saturday.)

To reproduce - try a project with a big sprite this: https://scratch.mit.edu/projects/107873291/#editor

1. Choose the saxophone sprite.
2. Open the Sound Library
3. Double-click the 'Music Loops' category
4. Click OK (or Cancel)
5. The cat sprite will be selected instead of the Sax (because it is underneath the area clicked in the library).

cwillisf:
Before: any click within the rectangle of the Scratch stage would be forwarded to the top-most sprite at that (x,y) location, if any, even if the sprite was covered by some other UI such as a media library window.

After: a click is forwarded to the top-most sprite at its (x,y) location if Flash reports the target of the click is either the Flash stage (this covers 3D mode) or some descendant of the Scratch stage (this covers 2D mode).

Testing:
1. Download [click-test.zip](https://github.com/LLK/scratch-flash/files/253114/click-test.zip) (rename to SB2) or make a project with:
   - At least two visually distinct backdrops
   - At least one sprite with a script like `when this sprite clicked: switch backdrop to {next backdrop}`
2. Click on the sprite, verifying that the backdrop cycles.
3. Click near the sprite, verifying that the backdrop does not cycle.
4. Click in menus, in the scripting area, in text fields within blocks, etc. to verify that other mouse input works as expected.
5. Add an effect block such as `change {color} effect by {25}` to the scripting area. This will cause the Scratch stage to enter 3D mode.
6. Repeat steps 2-4 in 3D mode.

This resolves #1100

### DIFF
--- a/src/util/GestureHandler.as
+++ b/src/util/GestureHandler.as
@@ -343,6 +343,15 @@ public class GestureHandler {
 		}
 	}
 
+	// Returns true if non-null `ancestor` is `descendant` or one of its ancestors.
+	private static function isAncestor(ancestor:DisplayObject, descendant:DisplayObject):Boolean {
+		while (descendant) {
+			if (ancestor == descendant) return true;
+			descendant = descendant.parent;
+		}
+		return false;
+	}
+
 	private function findMouseTarget(evt:MouseEvent, target:*):DisplayObject {
 		// Find the mouse target for the given event. Return null if no target found.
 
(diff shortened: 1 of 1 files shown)


src/util/GestureHandler.as
```diff
@@ -358,11 +380,7 @@ public class GestureHandler {
 			}
 			o = o.parent;
 		}
-		var rect:Rectangle = app.stageObj().getRect(stage);
-		if(!mouseTarget && rect.contains(evt.stageX, evt.stageY))  return findMouseTargetOnStage(evt.stageX / app.scaleX, evt.stageY / app.scaleY);
-		if (o == null) return null;
 		if ((o is Block) && Block(o).isEmbeddedInProcHat()) return o.parent;
-		if (o is ScratchObj) return findMouseTargetOnStage(evt.stageX / app.scaleX, evt.stageY / app.scaleY);
```
cwillisf, line 365: - The `if (!mouseTarget && rect.contains...` is now covered by the `checkScratchStage` logic above.
- So is the `if (o is ScratchObj)` case.
- The `if (o==null) return null` line is redundant given the `return o;` line below; it just short-circuits some `is` checking.

thisandagain:
Thanks for the thorough testing instructions. 👍

technoboy10:
Used the testing instructions on Chrome 50, Firefox 46, and Safari 9.1 on OS X 10.11.5.

- [x] Backdrop cycles only when clicking directly on sprite (2D and 3D mode)
- [x] Menus and text fields behave as expected

TheLogFather:
Note that it looks like this also fixes https://github.com/LLK/scratch-flash/issues/600. :)
(Had a feeling it might...)

cwillisf:
Thanks for pointing that out @TheLogFather!

PR #1108, opened 2016-05-09
Update the README to include Gradle instructions

### LINKED ISSUE #1046: README.md is out of date and missing some information
The README.md file in this repository should be updated:
- It should suggest using Gradle instead of Ant in order to simplify the SDK setup process.
- It should explain the differences between scratch-flash and what's found on scratch.mit.edu .

cwillisf:
This resolves #1046

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,17 +6,49 @@ If you're interested in contributing to Scratch, please take a look at the issue
 It's been said that the Scratch Team spends about one hour of design discussion for every pixel in Scratch, but some think that estimate is a little low. While we welcome suggestions for new features in our <a href="http://scratch.mit.edu/discuss/1/">suggestions forum</a> (especially ones that come with mockups), we are unlikely to accept PRs with new features that we haven't deeply thought through. Why? Because we have a strong belief in the value of keeping things simple for new users. To learn more about our design philosophy, see this <a href="http://scratch.mit.edu/discuss/post/1576/">forum post<a>, or <a href="http://web.media.mit.edu/~jmaloney/papers/ScratchLangAndEnvironment.pdf">this paper</a>.
 
 ### Building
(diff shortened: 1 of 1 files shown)


thisandagain:
This is great. Thanks @cwillisf

README.md
```diff
@@ -6,17 +6,49 @@ If you're interested in contributing to Scratch, please take a look at the issue
...
 
-The ``build.properties`` file sets the default location for the Flex SDK. Create a ``local.properties`` file to set the location on your filesystem. Your ``local.properties`` file may look something like this:
-```
-FLEX_HOME=/home/joe/downloads/flex_sdk_4.11
+There are two versions of the Scratch 2.0 editor that can be built from this repository. See the following table to determine the appropriate command for each version. When building on Windows, replace `./gradlew` with `.\gradlew`.
+
+Required Flash version | Features | Command
+--- | --- | ---
+11.6 or above | 3D-accelerated rendering | `./gradlew build -Ptarget=11.6`
+10.2 - 11.5 | Compatibility with older Flash (Linux, older OS X, etc.) | `./gradlew build -Ptarget=10.2`
```
Tymewalk, line 17: Quick note - I'm on Linux and I can target 11.6 just fine.
Not sure if I can target higher versions. What's the highest I can use with Scratch?

cwillisf, line 17: The "Required Flash version" column is referring to the Flash Player run-time requirements, not build-time requirements. The picture's a little complicated with Linux:
- You can probably target any Flash version when building on Linux, though I haven't tested this myself
- Chrome comes bundled with a current version of Flash, so as long as you're running Chrome you can run the Flash 11.6+ version of Scratch
- Outside of Chrome, the most recent standalone or plugin-based Flash Player available for Linux is version 11.2, which is not recent enough for 3D-accelerated rendering

PR #1116, opened 2016-05-23
    Added "Open in Scratch" functionality

bsb20:
```
-Allows projects in microworlds to be opened as remixes for the
currently logged in user
```


src/ui/parts/TopBarPart.as
```diff
@@ -37,7 +37,7 @@ import uiwidgets.*;
 
 public class TopBarPart extends UIPart {
 
-	private var shape:Shape;
+	protected var shape:Shape;
```
cwillisf: Is this change necessary? I don't see it being used here or in the other PR...

PR #1119, opened 2016-06-01
Make "point towards" default to "mouse-pointer"

### LINKED ISSUE #1087: "point towards" Motion block defaults to blank rather than mouse-pointer
Scratch: v445.2
scratch-flash: 2a019c6
scratch-flash-online: 015f4cc
Browser: Chrome Version 50.0.2661.86 (64-bit)

Steps to repro:
- In the editor (click Create button to get there) >
- With a Sprite active (Scratch cat is active by default) >
- In the Motion blocks section of the Script tab >
- Note that the "point towards" block has a blank value for what is being pointed towards.
- Drag the block into the Script area to the right.
- Click on the small black arrow in the block to show the drop-down menu.
- Note that there is only 1 option in the drop-down, which is "mouse-pointer".
- Choose "mouse-pointer".
- Click on the small black arrow in the block to show the drop-down menu again.
- Note that there is still only 1 option, "mouse-pointer", so you cannot re-select the blank option that showed up in the block initially.
- If you refresh the page and blow away your changes, you will see the blank option reappear in the block.
(- Adding more sprites will add more options to the drop-down, you can choose the sprites from the drop-down.)

If you look at the "go to [mouse-pointer]" block, you can see it defaults to "mouse-pointer", so that might be what this "point towards" block should default to as well.

screen shot 2016-04-22 at 10 48 55 am
screen shot 2016-04-22 at 10 51 07 am
screen shot 2016-04-22 at 10 51 15 am
screen shot 2016-04-22 at 10 51 33 am

cwillisf:
This resolves #1087


src/Specs.as
```diff
@@ -117,10 +117,10 @@ public class Specs {
 		["turn @turnLeft %n degrees",			" ", 1, "turnLeft:",				15],
 		["--"],
 		["point in direction %d.direction",		" ", 1, "heading:",					90],
-		["point towards %m.spriteOrMouse",		" ", 1, "pointTowards:",			""],
+		["point towards %m.spriteOrMouse",		" ", 1, "pointTowards:",			"mouse-pointer"],
 		["--"],
 		["go to x:%n y:%n",						" ", 1, "gotoX:y:"],
-		["go to %m.location",				" ", 1, "gotoSpriteOrMouse:",		"mouse-pointer"],
+		["go to %m.location",					" ", 1, "gotoSpriteOrMouse:",		"mouse-pointer"],
```
cwillisf, line 123: This change makes more sense when tab = 4 spaces.
Try adding `?ts=4` to the end of this page's URL...

jwzimmer:
**Test cases:**
- [ ] The block should still exist in the Motion section
- [ ] The "point towards [ ]" block should default to "point towards [mouse-pointer]"
- [ ] The block should be able to be dragged into the scripting area
- [ ] If you create more than 1 sprite, the other sprites should show up in the menu of "point towards" options
- [ ] Clicking on the "point towards" block when it's in the script area should cause the sprite to point towards the chosen option
  - [ ] for mouse-pointer
  - [ ] for a sprite
- [ ] Using the "point towards" block in a script should work as expected
  - [ ] for mouse-pointer
  - [ ] for a sprite

Issue #1122, opened 2016-06-03
ArgumentError: Error #2015 - Project gets stuck on fullscreen view, no content, but tips window appears at the right


jwzimmer:
In the JavaScript Console, @cwillisf and I both saw Error # 2015 repeated a lot of times (as long as the page was open, more error 2015's would appear).
On Sentry: https://app.getsentry.com/scratch-team/scratch-flash/issues/105167534/

In freshdesk ticket 49353, a user reported that their project does not load properly:

> The untitled 15 project is not loading properly. Can you please have a look at it as there was a long script and it now says it's empty.

Project ID is 108737817 (shared project).

Repro:
- Sign in as an admin
- Go to the project page directly via URL
- OR Go to the user's My Stuff page
- Click on the "See Inside" button for that project (Untitled 15)
- The project will load with a grey background and the tips window to the right side. You can open and close the tips window. The play and pause buttons don't seem to do anything.

I saw the project fail to load in Chrome Version 50.0.2661.102 (64-bit) (on a Mac). @cwillisf was able to repro the issue in IE 11 (on a Windows machine). The project loaded normally when I tried it in Safari (on a Mac) and when @cwillisf tried it in Edge and Chrome (on a Windows machine).

<img width="268" alt="screen shot 2016-06-03 at 2 56 48 pm" src="https://cloud.githubusercontent.com/assets/1320456/15790022/19a3fe5e-299d-11e6-8024-19ea796e5c84.png">

jwzimmer:
I'm seeing this issue with another project, where the project gets stuck in full-screen mode with the tips window at the side, and otherwise won't load.
Project ID: 110776758

jwzimmer:
@cwillisf Seems to be working now. I cannot repro in...
- Chrome 52 on Mac OS X El Capitan
- IE 11 on Win 7 VM
- Firefox 48 on Win 7 VM (although the editor is too small - known issue)
- Chrome 52 on Win 7 VM
- Firefox 48 on Mac OS X El Capitan

yania:
I have the same problem with this Project 
https://scratch.mit.edu/projects/147511896/
In Windows 10, checked with Chrome and with Edge.
It is my son'sproject. The Project Works and suddenly start getting stucked this way.

yania:
11:35:17 PM | err | TypeError: Error #1009
eval code (224) (1,7)
11:35:17 PM | err | TypeError: Error #1009
eval code (224) (1,7)
[...]

PR #1127, opened 2016-06-06
Limit max size in RasterHull()

cwillisf:
`RasterHull()` now limits the size of the `BitmapData` which it creates. The limit is the `texSizeMax` from `DisplayObjectContainerIn3D`, which is now public. Prior to this change the `BitmapData` size was not limited, which could cause an exception to be thrown by the `BitmapData` constructor.

To test:
- Open the browser console
- Load a project which exhibits this problem: https://app.getsentry.com/scratch-team/scratch-flash/issues/105167534/
  (or contact me for an SB2 which triggers the bug)

This is one of at least two issues which cause the symptom described in #1122

### DIFF
--- a/src/render3d/DisplayObjectContainerIn3D.as
+++ b/src/render3d/DisplayObjectContainerIn3D.as
@@ -26,7 +26,11 @@ package render3d {
 
 import flash.display.Sprite;
 
-public class DisplayObjectContainerIn3D extends Sprite {SCRATCH::allow3d{
+public class DisplayObjectContainerIn3D extends Sprite {
+	public static var texSizeMax:int = 2048;
+	public static var texSize:int = 1024;
+	public static var maxTextures:uint = 15;
+SCRATCH::allow3d{
 	import com.adobe.utils.*;
 
 	import filters.FilterPack;
@@ -86,9 +90,6 @@ public class DisplayObjectContainerIn3D extends Sprite {SCRATCH::allow3d{
 	private var textures:Array;
 	private var testBMs:Array;
 	private var textureIndexByID:Object;
-	private static var texSizeMax:int = 2048;
-	private static var texSize:int = 1024;
-	private static var maxTextures:uint = 15;
 	private var penPacked:Boolean;
 
 	/** Triangle index data */
--- a/src/scratch/ScratchCostume.as
+++ b/src/scratch/ScratchCostume.as
@@ -305,7 +305,7 @@ public class ScratchCostume {
 		var s:Shape = shapeDict[id];
 		if (!s) {
 			s = new Shape();
-			var pts:Array = RasterHull();
+			var pts:Vector.<Point> = RasterHull();
 			s.graphics.clear();
 
 			if (pts.length) {
@@ -335,12 +335,14 @@ public class ScratchCostume {
 	 object is composed of
 	 ** a single point
 	 */
-	private function RasterHull():Array {
+	private function RasterHull():Vector.<Point> {
+		var H:Vector.<Point> = new Vector.<Point>();
 		var dispObj:DisplayObject = displayObj();
 		var r:Rectangle = dispObj.getBounds(dispObj);
-//trace('flash bounds: '+r);
-		if (r.width < 1 || r.height < 1)
-			return [new Point()];
+		if (r.width < 1 || r.height < 1) {
+			H.push(new Point());
+			return H;
+		}
 
 		r.width += Math.floor(r.left) - r.left;
 		r.left = Math.floor(r.left);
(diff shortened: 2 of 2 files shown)


src/scratch/ScratchCostume.as
```diff
@@ -344,9 +346,16 @@ public class ScratchCostume {
...
+		var desiredWidth: int = Math.max(0, Math.ceil(r.width));
+		var desiredHeight: int = Math.max(0, Math.ceil(r.height));
+		if (desiredWidth >= DisplayObjectContainerIn3D.texSizeMax
+				|| desiredHeight >= DisplayObjectContainerIn3D.texSizeMax) {
+			var factor:Number = (DisplayObjectContainerIn3D.texSizeMax - 1) / Math.max(desiredWidth, desiredHeight);
+			desiredWidth *= factor;
+			desiredHeight *= factor;
+		}
+		// TODO: figure out why we add 1 to each dimension in addition to using Math.ceil above
+		var image:BitmapData = new BitmapData(desiredWidth + 1, desiredHeight + 1, true, 0);
```
cwillisf, line 360: Luckily the code below is already set up to use whatever size `image` happens to get here.

src/scratch/ScratchCostume.as
```diff
@@ -344,9 +346,16 @@ public class ScratchCostume {
...
-		var image:BitmapData = new BitmapData(
-				Math.max(1, Math.ceil(r.width) + 1), Math.max(
-						1, Math.ceil(r.height) + 1), true, 0);
+		var desiredWidth: int = Math.max(0, Math.ceil(r.width));
+		var desiredHeight: int = Math.max(0, Math.ceil(r.height));
+		if (desiredWidth >= DisplayObjectContainerIn3D.texSizeMax
+				|| desiredHeight >= DisplayObjectContainerIn3D.texSizeMax) {
+			var factor:Number = (DisplayObjectContainerIn3D.texSizeMax - 1) / Math.max(desiredWidth, desiredHeight);
+			desiredWidth *= factor;
+			desiredHeight *= factor;
```
cwillisf, line 357: scale while maintaining aspect ratio

src/scratch/ScratchCostume.as
```diff
@@ -344,9 +346,16 @@ public class ScratchCostume {
...
+		}
+		// TODO: figure out why we add 1 to each dimension in addition to using Math.ceil above
+		var image:BitmapData = new BitmapData(desiredWidth + 1, desiredHeight + 1, true, 0);
 //trace('bitmap rect: '+image.rect);
```
thisandagain: Any idea why this (and some below) commented out code was left in here?

src/render3d/DisplayObjectContainerIn3D.as
```diff
@@ -26,7 +26,11 @@ package render3d {
 
 import flash.display.Sprite;
 
-public class DisplayObjectContainerIn3D extends Sprite {SCRATCH::allow3d{
+public class DisplayObjectContainerIn3D extends Sprite {
+	public static var texSizeMax:int = 2048;
```
thisandagain, line 30: How did you determine these limits?

cwillisf, line 30: A while back we allowed up to 4096x4096, but that was causing a lot of noise on Sentry due to systems which didn't support it. The 2048x2048 limit was put in place by @sclements in #897. As he points out in that PR, at that time the top event was "Texture too big (max is 2048x2048)"

There doesn't appear to be any way to query Flash about maximum texture size, but we could increase texture size in a loop until we catch a texture-too-big exception.

Of course, all that is talking about 3D hardware. The change in this PR is important even when 3D acceleration isn't available. I use `texSizeMax` because a) it seems like a reasonable limit and b) it makes sense as a limit when we actually are running in 3D mode.

src/scratch/ScratchCostume.as
cwillisf: Good point -- I'll remove it.

TheLogFather:
Is it this change which broke this platformer: https://scratch.mit.edu/projects/104297197/ ...?
(See https://scratch.mit.edu/discuss/topic/208322)

cwillisf:
@TheLogFather yes, this change probably did cause that project to break. Unfortunately there's no good answer: prior to this change, that project would cause the editor to crash on some computers. Now it doesn't crash on those computers, but it doesn't work right on any computer.

We could adjust the limit according to what each computer is capable of handling, but that would mean projects like this would work on some computers but fail on others. I think that would feel like a bug.

In my opinion it's best to have consistent behavior across as many computers as we reasonably can, and unfortunately right now that means limited support for costumes above 2048 in either dimension.

TheLogFather:
Yeah, I suspected as much (hence my recommendation).

PR #1130, opened 2016-06-08
Allow experimental extensions from coding.me

### LINKED ISSUE #1128: Add `*.coding.me` to list of allowed domains for extensions
`github.com` and `github.io` are blocked in China which results in Scratch Extensions not being available to a large number of our users. To work around this, we have a partner that would like to start porting some extensions over to Mandarin Chinese and hosting them via `coding.net` and their equivalent to `github.io` which is called `coding.me`.

[ExtensionManager.as](https://github.com/LLK/scratch-flash/blob/17e095682939f59163ee0a4806a71aa8603267fd/src/extensions/ExtensionManager.as)

/cc @cwillisf @kaschm

cwillisf:
This resolves #1128

Testing:
- Download each of these projects, renaming the files to `*.sbx`
  - [forbidden.zip](https://github.com/LLK/scratch-flash/files/303627/forbidden.zip)
  - [mesh@github.zip](https://github.com/LLK/scratch-flash/files/303628/mesh.github.zip)
- Load each project.

Expected results:
- `mesh@github.sbx` attempts to load an extension hosted on `github.io`, which should succeed: you should see the Mesh blocks in the "More Blocks" category.
- `forbidden.sbx` attempts to load an extension hosted on `forbidden.url`, which should fail in the console with the error message below.

Note that, since the site `forbidden.url` doesn't exist, the "forbidden" extension will always fail to load. Testing should only pass if the message below appears in the JS console.

> Refusing to load project extension "Mesh": Experimental extensions must be hosted on an approved domain. Approved domains are: .github.io, .coding.me

### DIFF
--- a/src/extensions/ExtensionManager.as
+++ b/src/extensions/ExtensionManager.as
@@ -55,6 +55,13 @@ public class ExtensionManager {
 	static public const wedoExt:String = 'LEGO WeDo';
 	static public const wedo2Ext:String = 'LEGO WeDo 2.0';
 
+	// Experimental extensions must be hosted on one of these domains
+	// These should start with '.' to avoid accepting things like 'malicious.not_github.io'
+	static public const allowedDomains:Vector.<String> = new <String>[
+			'.github.io',
+			'.coding.me'
+	];
+
 	public function ExtensionManager(app:Scratch) {
 		this.app = app;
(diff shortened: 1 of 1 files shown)


thisandagain:
👍  Thanks.

PR #1147, opened 2016-06-20
Store bitmap resolution in backpack

cwillisf:
Along with a corresponding change in scratch-flash-online, this fixes backdrops "zooming in" in the backpack.
Also, make a class constant for the special value which causes ScratchCostume to calculate a default rotation center for the costume.

Testing:
- Backdrops in the backpack should no longer "zoom in":
  1. Add a backdrop to your backpack.
  2. Ensure that the thumbnail looks correct.
  3. Reload the project page
  4. Ensure that the thumbnail still looks correct.
- Costumes and backdrops outside of the backpack should not be affected in any way
  - The rotation center of sprites should still work as expected
- Other backpack items should not be affected in any way

### DIFF
--- a/src/ui/media/MediaInfo.as
+++ b/src/ui/media/MediaInfo.as
@@ -54,6 +54,7 @@ public class MediaInfo extends Sprite {
 	public var objType:String = 'unknown';
 	public var objName:String = '';
 	public var objWidth:int = 0;
+	public var bitmapResolution:int = 1;
 	public var md5:String;
 
 	public var owner:ScratchObj; // object owning a sound or costume in MediaPane; null for other cases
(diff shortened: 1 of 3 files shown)


src/ui/media/MediaInfo.as
```diff
@@ -291,6 +293,9 @@ public class MediaInfo extends Sprite {
 		if (mycostume) {
 			result.width = mycostume.width();
 			result.height = mycostume.height();
+			if (mycostume.bitmapResolution != 1) {
+				result.bitmapResolution = mycostume.bitmapResolution;
+			}
```
cwillisf, line 298: The `if` is here to shrink the JSON a bit when we don't need to store the resolution.

PR #1156, opened 2016-06-29
Delete Unnecessary Whitespace

bjskistad:
I don't know if this change is necessary, or if the whitespace is needed, but I created a PR just in case.


cwillisf:
Hi @bjskistad! The reason the text doesn't line up properly in some contexts is that the team uses a tab width of 4 whereas GitHub and some other editors use a tab width of 8. If you set your editor to a tab width of 4, this text should line up correctly.

In general I'd prefer to avoid changes that are purely related to formatting unless it's important user-facing text or something like that. That said, I appreciate the attention to detail!

PR #1157, opened 2016-07-01
Keep parens on translation strings by default

cwillisf:
Previously, the translation string export process would remove parentheticals from all strings by default. There was a way to request that this should be skipped, but no callers used it.

Now, only one specific caller explicitly requests that parentheticals should be stripped and all other callers have their strings passed unmodified. This one specific caller is the only case where we actually expect parentheticals to be stripped.

### DIFF
--- a/src/uiwidgets/Menu.as
+++ b/src/uiwidgets/Menu.as
@@ -77,7 +77,8 @@ public class Menu extends Sprite {
 	public function showOnStage(stage:Stage, x:int = -1, y:int = -1):void {
 		if (stringCollectionMode) {
 			for each (var item:MenuItem in allItems) {
-				TranslatableStrings.add(item.getLabel());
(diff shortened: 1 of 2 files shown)


src/uiwidgets/Menu.as
```diff
@@ -77,7 +77,8 @@ public class Menu extends Sprite {
 	public function showOnStage(stage:Stage, x:int = -1, y:int = -1):void {
 		if (stringCollectionMode) {
 			for each (var item:MenuItem in allItems) {
-				TranslatableStrings.add(item.getLabel());
+				// TODO: do we really want to remove parentheticals on all these?
+				TranslatableStrings.add(item.getLabel(), true);
```
cwillisf, line 81: This is for cases like the drum and instrument list, where menu items have the form `(42) Foo`.

thisandagain:
Simple enough. Thanks!

PR #1179, opened 2016-08-02
Trying to fix #1166 & related stuff

### LINKED ISSUE #1166: #1119 should initialise to _mouse_, not mouse-pointer
This means that the "point towards mouse-pointer" block is not working when first dragged out from the block palette, until the menu item is selected again.

For example, see https://scratch.mit.edu/discuss/topic/208457/ and probably also https://scratch.mit.edu/discuss/topic/208622/
Was also reported in Bugs & Glitches forum here: https://scratch.mit.edu/discuss/topic/208762/

EDIT: The title doesn't create a reference, so adding this: #1119

joker314:
Fixes #1166 I think and I am not sure if I did it right? (A bug that means the point towards mouse-pointer block does not work, **preventing some users from understanding what it does**)

This also adds the default drop-downs to the sensing blocks "touching?" and "distance to"

And, it removes a line in ScratchSprite.as because I changed Specs.as to make it not needed

EDIT: Untested- I can't be bothered to download and build it, sorry


TheLogFather:
Yes, that'll do the trick.

Although there's a strange inconsistency, since the "go to" block also sets itself up wrongly too (as you noticed), but that gets fixed by the line in defaultArgsFor (ScratchSprite.as) which sets up its default value as "_mouse_".  :/

I guess the question is which way the ST want to do this – through the value in Specs.as, or through defaultArgsFor? (Though I don't really see the point of setting them to an incorrect value in Specs.as...)

TheLogFather:
(And, while we're sorting out drop-down menu initial states, can we not also fix the ones that are missing for the "touching" and "distance to" blocks?)

joker314:
So if that function fixes goTo, does that mean that changing the value here neither harms  nor helps the goTo block, right? Since it gets fixed anyway?

As for the other dropdowns, I'll add them now

joker314:
Ok, I think I did that right: so now those blocks should have (correct) default dropdowns

TheLogFather:
Yes, looks fine.

Yes, changing the value here for the "go to" block makes no difference since the value in defaultArgsFor will override it.

However, I don't really see the point of it setting the default value for that block there. The purpose of that function seems to be to initialise blocks in the block palette that have entries which depend on the current state of the sprite (e.g. the "go to X Y" block in the block palette shows the current X,Y position). it doesn't seem to be meant for blocks that always have a fixed initial entry.

Myself, I'd probably 'sanitise' defaultArgsFor by removing the set up of the "go to" block's menu. Also, I wonder if the video motion block setting also could be removed (and initialised in Specs.as) – though I've not checked if it has the correct initial settings in Specs.as (but could be fixed if not), nor if there is some special circumstance that means it does need overriding in defaultArgsFor.

However, it's up to the ST how they'd like to proceed with initialising these things, so...

TheLogFather:
(OK... answering my own question... yes, the video block does need special treatment, otherwise it'll initialise with a blank menu entry in a sprite's block palette [but it's ok for Stage], due to the extra block spec for that in Specs.as. So scrap that particular idea...)

joker314:
I'll (try) to remove the line  in the function, that is, if I can use GitHub well enough..

joker314:
So to change multiple files I have to use git. Oh well.. I'll do that then

joker314:
So, it is currently telling me I don't have push rights to the repo. I must have forked the master thing or something?

joker314:
Hmm, well, until I learn to use Git properly, I think I'll leave this PR as that for now.

TheLogFather:
The way to do it is to create your own repo on github, as a fork of the original, and then you can push to that. You can then make a pull request from there. (I'd recommend that you create a branch for each different PR that you might be interested in making.)
As an example, here's mine (not done much to it recently): https://github.com/TheLogFather/scratch-flash
Note that I have a branch for different fixes (e.g. https://github.com/TheLogFather/scratch-flash/tree/avoid-checkblockingargs and https://github.com/TheLogFather/scratch-flash/tree/fix-3d-brightness).

joker314:
Ah, I see. I thought a branch was another word for a fork, but I take it now that it isn't. Ok, I'll do my best :)

joker314:
Oh I did it I think! Yey!

joker314:
Great! Thanks so much! This was my first PR to Scratch

PR #1212, opened 2016-10-12
Updgrade Gradle builds to apache-flex-sdk 4.15.0

### LINKED ISSUE #1195: Build Failure Because of Swfobject
http://swfobject.googlecode.com/files/swfobject_2_2.zip is a 404 link, since google stopped support for googlecode a while ago. This is causing build errors because the link above is a 404 link. The error is most likely caused by the downloads.xml file in apache flex's zip.

cwillisf:
This shouldn't have any noticeable impact on the editor, and I haven't noticed anything in my testing, but we should consider giving the editor a more thorough check than usual before deploying this build.

This resolves #1195

### DIFF
--- a/scratch.gradle
+++ b/scratch.gradle
@@ -42,7 +42,7 @@ def scratchFlashCommitID = getCommitID(commonDir)
 println "Commit ID for scratch-flash is: ${scratchFlashCommitID}"
 
 dependencies {
-    flexSDK group: 'org.apache', name: 'apache-flex-sdk', version: '4.10.0', ext: 'zip'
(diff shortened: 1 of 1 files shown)


thisandagain:
👍

rschamp:
I played around in the editor for a bit and found the sounds tab was not working. Not sure if that's related to this or not, but logged it here: https://github.com/LLK/scratch-flash-online/issues/102

rschamp:
Since the sounds tab was found to be a result of https://github.com/LLK/scratch-flash/pull/1214 removing blocker from this.

PR #1214, opened 2016-10-14
Speculative fixes for sounds being replaced by empty sound

### LINKED ISSUE #1193: User reported sound clips disappearing from projects
Reported via freshdesk no. 57005:

> Hello there is a major glitch going around on scratch that i have noticed you see many people make music playing projects however these projects will sometimes have a 4 minute long sound file and yet have no audio the weird thing is is that so far ive only noticed it on music playing projects in fact i even made one and it was good for about a week before half the songs dissapeared this is it here 118351985 what is causing this and how is it fixed? 

Metadata: 
- User: (see project page or freshdesk ticket)
- Operating System: Windows 7 
- Browser: Firefox 48.0 
- Flash Version: Flash 22.0 (release 0)

**I'm wondering if the same issue that caused these sound clips to go missing is also what caused the missing sound clips from https://github.com/LLK/scratch-flash/issues/1110#issuecomment-242832383.**

cwillisf:
Some users have reported valid sounds being replaced with an empty sound while working with the online editor. I have not been able to reproduce this behavior, but based on the evidence we've collected so far I made some guesses about what might be happening.

The changes here include:
- General cleanup & reformatting of the code in `ScratchSound.as`, including the removal of some long-unused code.
- Turn the `md5` property into a read-only property that is calculated on demand, cached, and invalidated when the sound data changes. This was already partially implemented, but now it's complete.
- Fix a case where a sound could arbitrarily get `soundID=0` even if that's the valid ID of another sound.
- Add some logging to error cases, which might help diagnose the problem if these changes don't fix it.

Note that the logging is currently set to use `LogLevel.WARNING` which means these events will show up in a browser console but won't be logged in Sentry. I could bump some of these new logging events to `LogLevel.ERROR` if we want to see them in Sentry, but I'm nervous about overwhelming our event quota.

Hopefully this fixes #1193

### DIFF
--- a/src/scratch/ScratchSound.as
+++ b/src/scratch/ScratchSound.as
@@ -118,7 +118,6 @@ public class ScratchSound {
 				samples = (channels == 2) ? stereoToMono(samples, (newRate < rate)) : downsample(samples);
 			}
 			setSamples(samples, newRate, true);
-			soundID = 0;
 		}
 		else if ((soundData.length > compressionThreshold) && ('' == format)) {
 			// Compress large, uncompressed sounds
--- a/src/scratch/ScratchSound.as
+++ b/src/scratch/ScratchSound.as
@@ -31,7 +31,6 @@
 package scratch {
 import by.blooddy.crypto.MD5;
 
-import flash.media.Sound;
 import flash.utils.*;
 
 import logging.LogLevel;
@@ -39,14 +38,11 @@ import logging.LogLevel;
 import sound.*;
 import sound.mp3.MP3Loader;
 
-import util.*;
-
 public class ScratchSound {
 
 	public var soundName:String = '';
 	public var soundID:int;
 	public var md5:String;
-	private var __soundData:ByteArray = new ByteArray();
 	public var format:String = '';
 	public var rate:int = 44100;
 	public var sampleCount:int;
@@ -55,20 +51,28 @@ public class ScratchSound {
 
 	public var editorData:Object; // cache of data used by sound editor; not saved
 	public var channels:uint = 1;
-	private const WasEdited:int = -10; // special soundID used to indicate sounds that have been edited
-	SCRATCH::allow3d
-	public var nativeSound:Sound;
 
-	// Undo support; not saved
-	public var undoList:Array = [];
-	public var undoListIndex:int;
+	public function get soundData():ByteArray {
+		return __soundData;
+	}
+
+	public function set soundData(data:ByteArray):void {
+		__soundData = data;
+		md5 = null;
+	}
+
+	private const WasEdited:int = -10; // special soundID used to indicate sounds that have been edited
+
+	private var __soundData:ByteArray = new ByteArray();
 
 	public function ScratchSound(name:String, sndData:ByteArray) {
 		this.soundName = name;
 		if (sndData != null) {
 			try {
 				var info:Object = WAVFile.decode(sndData);
-				if ([1, 3, 17].indexOf(info.encoding) == -1) throw Error('Unsupported WAV format');
+				if ([1, 3, 17].indexOf(info.encoding) == -1) {
+					throw Error('Unsupported WAV format');
+				}
 				soundData = sndData;
 				if (info.encoding == 17)
 					format = 'adpcm';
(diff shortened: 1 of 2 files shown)


src/scratch/ScratchSound.as
```diff
@@ -81,40 +92,40 @@ public class ScratchSound {
...
 			var samples:Vector.<int> = WAVFile.extractSamples(soundData);
-			if (rate > 32000 || channels == 2)
-				samples = (channels == 2) ?
-					stereoToMono(samples, (newRate < rate)) :
-					downsample(samples);
+			if (rate > 32000 || channels == 2) {
+				samples = (channels == 2) ? stereoToMono(samples, (newRate < rate)) : downsample(samples);
+			}
 			setSamples(samples, newRate, true);
-			soundID = 0;
```
cwillisf, line 110: Sound IDs are array indices so zero is a valid sound ID and could refer to some other sound. My guess is that the intent was to invalidate the sound ID, but `setSamples` already does that so I believe it's safe to remove this line.

thisandagain:
Thanks @cwillisf. This looks good. I think keeping the log level as `WARNING` makes sense given our Sentry quotas. @jwzimmer this means that should we receive helpdesk tickets about this issue in the future we'll need to ask users to look at their browser console. Does that seem reasonable to you?

jwzimmer:
@thisandagain & @cwillisf, yes, asking helpdesk users reporting this issue to look in the console sounds reasonable to me, thanks.

rschamp:
Oh, this is more likely the culprit for what I reported at LLK/scratch-flash-online#102

jwzimmer:
**Checking that https://github.com/LLK/scratch-flash-online/issues/102 is resolved in the recent build e590824.**

- I can load the sounds tab in https://scratch.ly/projects/1000014188/#editor
- However I see a relevant-looking warning in the console:
screen shot 2016-10-31 at 10 14 24 am
- The warning doesn't match any of the warnings I see that @cwillisf added for this PR. I'm not sure if that means this is not relevant (a failure we already know about and handle) or if it's something to revisit here.
- When I look in the changes @cwillisf made, I see that getSamples() and reduceSizeIfNeeded() call WAVFile.extractSamples(soundData), which seems potentially related to the warning I'm seeing?

rschamp:
@jwzimmer The "Error extracting samples from WAV file" warning looks like the fix @cwillisf added (this PR introduced a regression that he fixed with #1225). So I think this is a good thing and can be considered resolved.

jwzimmer:
@rschamp Great, thanks. 👍

jwzimmer:
QA for this PR (https://github.com/LLK/scratch-flash/pull/1214#issuecomment-257306869 was specifically to test that the new build resolved the blocking issue with the sounds tab not loading).

Firefox 49.0.2 on Mac OS 10.11
- While signed out, clicked Create
- Deleted Meow default sound
- Added a sound from library
- Tried to upload a random file as a sound (got error in dialogue box, which is good)
  > Error while constructing sound:WAVFile: bad file header Object { info: null, exception: Object }
- Recorded a lot of "blank" sounds (~30)
- Recorded a real sound
- Replayed the sound, worked as expected
- Signed in as jwzimmer1990
- Navigated away from project
- Returned to project via My Stuff
- The sounds seem to be as expected
- I played the sound I recorded, it worked as expected
- Deleted some sounds
- Added "when green flag clicked, play recorded sound" script
- Clicked block stack, sound played as expected
- From project page, green flag played the sound as expected
- Added a blank sound and recorded sound to 1000014117 (an older project)
- Added "when green flag clicked, play recorded sound" script
- Clicked block stack, sound played as expected
- From project page, green flag played the sound as expected
- Returned to the 1st project (1000014233)
- Sounds seemed as expected
- Could play recorded sound
- Returned to 2nd project
- Sounds seemed as expected
- Could play recorded sound

cwillisf:
@jwzimmer & @rschamp, thanks for helping on this! @jwzimmer double-thanks for such a thorough and well-documented set of test cases!

The "Error extracting samples from WAV file" message was indeed added with this pull request. If you see that message you're probably in a situation that would have triggered a blank "Sounds" tab prior to this change.

PR #1217, opened 2016-10-17
Fix Boolean reporters in extensions when offline

### LINKED ISSUE #1117: extension boolean reporters not working offline
@cwillisf I'm not having any luck with the new Wedo2 tilt boolean reporter in 446 offline - should 'b' blocks get the isRequester treatment too?

https://github.com/LLK/scratch-flash/blob/a61593e5d7fbb17a243643c9fc769490d5aaaf9e/src/blocks/Block.as#L134

cwillisf:
Extension reporters must run asynchronously in the offline editor. This change makes Boolean reporters act like requesters in the offline editor, just like we do with non-Boolean reporters.

Note that we don't normally support Boolean requesters -- there's no `'B'` block type -- but the code to handle extension requesters doesn't care whether the block is Boolean or not so this change is sufficient. It would be trivial to add support for `'B'` now.

This resolves #1117

### DIFF
--- a/src/blocks/Block.as
+++ b/src/blocks/Block.as
@@ -134,6 +134,8 @@ public class Block extends Sprite {
 		} else if (type == "b") {
 			base = new BlockShape(BlockShape.BooleanShape, color);
 			isReporter = true;
+			forceAsync = Scratch.app.extensionManager.shouldForceAsync(op);
+			isRequester = forceAsync;
 			indentLeft = 9;
 			indentRight = 7;
 		} else if (type == "r" || type == "R") {


cwillisf:
(this was a lot easier than I was expecting...)

src/blocks/Block.as
```diff
@@ -134,6 +134,8 @@ public class Block extends Sprite {
 		} else if (type == "b") {
 			base = new BlockShape(BlockShape.BooleanShape, color);
 			isReporter = true;
+			forceAsync = Scratch.app.extensionManager.shouldForceAsync(op);
+			isRequester = forceAsync;
```
cwillisf, line 138: Compare to the handling for `'r'` blocks below, which are the same as `'b'` other than the return type and the graphical shape of the block.

thisandagain:
I thought for sure this was going to be a rabbit hole. That's great. Thanks.

rschamp:
Looks good to me!

tmickel (approved):

PR #1222, opened 2016-10-20
Separate presentation mode request & reaction

cwillisf:
For better or worse, the editor tracks "presentation mode" (full-screen) state in the state of the presentation mode toggle button. This is fine as long as that button is the only way to change presentation mode, but sometimes we want JavaScript to set the presentation mode state. This change separates the concept of requesting a mode change (such as when JS requests a state) from the reaction to a change (such as when the button is clicked, or the JS request results in a change).

This is part of a fix for LLK/scratchr2#3639
See also:
- LLK/scratchr2#4173
- LLK/scratch-flash-online#288

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -628,6 +628,12 @@ public class Scratch extends Sprite {
 	protected var wasEditing:Boolean;
 
 	public function setPresentationMode(enterPresentation:Boolean):void {
+		if (stagePart.isInPresentationMode() != enterPresentation) {
+			presentationModeWasChanged(enterPresentation);
+		}
+	}
+
+	public function presentationModeWasChanged(enterPresentation:Boolean):void {
 		if (enterPresentation) {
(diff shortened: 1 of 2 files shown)


src/Scratch.as
```diff
@@ -646,6 +652,7 @@ public class Scratch extends Sprite {
 		for each (var o:ScratchObj in stagePane.allObjects()) o.applyFilters();
 
 		if (lp) fixLoadProgressLayout();
+		stagePart.presentationModeWasChanged(enterPresentation);
```
rschamp, line 655: Why does this not cause an infinite loop?

cwillisf: When `stagePart.presentationModeWasChanged` calls `fullscreenButton.setOn(isPresentationMode)`, that doesn't trigger the button's callback.

cwillisf: (is that the loop you were worried about?)

Issue #102, opened 2014-05-12
Improve dragging and dropping of media into the editor


scratchteam:
<a href="https://github.com/idreyn"><img src="https://avatars.githubusercontent.com/u/2208769?" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [idreyn](https://github.com/idreyn)**
_Sunday Feb 02, 2014 at 19:49 GMT_
_Originally opened as https://github.com/LLK/scratch-flash-common/issues/1323_

---

There is some support but it needs improvement to support more browsers.
**Reporter:** Shane Clements
**Assigned to:** Shane Clements
**Begin:** 2013-12-12
**Completed:** 0

PR #1225, opened 2016-10-27
Fix scratch-flash-online #102

cwillisf:
There are two changes here; see commits for detail. Summary:
- Fix a problem loading sounds, introduced by #1214 
- Catch and log errors during WAV decode rather than breaking the Sounds tab as described in LLK/scratch-flash-online#102

Testing:
- Make a new project.
- Save the project and reload it.
- Ensure that both sounds loaded correctly ("meow" on the sprite and "pop" on the backdrop).
- Add a new sound by clicking the "Record a new sound" button. Leave it empty.
- Save the project and reload it, then double-check that all three sounds are expected.
- Save the project to an SB2 and load that SB2 into the offline editor.
- Ensure that all three sounds loaded correctly in the offline editor.

This resolves LLK/scratch-flash-online#102 in its current sound-related form.

### DIFF
--- a/src/scratch/ScratchSound.as
+++ b/src/scratch/ScratchSound.as
@@ -76,6 +76,7 @@ public class ScratchSound {
 
 	public function ScratchSound(name:String, sndData:ByteArray) {
 		this.soundName = name;
+		this.soundID = WasEdited;
 		if (sndData != null) {
 			try {
 				var info:Object = WAVFile.decode(sndData);
@@ -242,11 +243,7 @@ public class ScratchSound {
 			bitsPerSample = 4;
 		}
 		reduceSizeIfNeeded(1); // downsample or compress to reduce size before saving
-		if (soundID == WasEdited) {
-			// sound was edited; force md5 to be recomputed
-			__md5 = null;
-			soundID = -1;
-		}
+		soundID = -1;
 	}
 
 	public static function isWAV(data:ByteArray):Boolean {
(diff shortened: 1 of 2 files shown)


src/scratch/ScratchSound.as
```diff
@@ -242,11 +243,7 @@ public class ScratchSound {
 			bitsPerSample = 4;
 		}
 		reduceSizeIfNeeded(1); // downsample or compress to reduce size before saving
-		if (soundID == WasEdited) {
-			// sound was edited; force md5 to be recomputed
-			__md5 = null;
-			soundID = -1;
-		}
+		soundID = -1;
```
cwillisf, line 246: I made this change after comparing this code to similar code dealing with images. That code is approximately:

``` js
if (imageID == WasEdited) { md5 = null; }
imageID = -1;
```

Clearing the MD5 isn't necessary here because it's automatically cleared at the same time the ID becomes `WasEdited`, so I just always set `soundID = -1;`.

Why is `soundID = -1` important? Copied from 86be65b284951f17b0a0e44adf238fd7b62adbfa's commit message:

> Additionally, the sound ID is now always set to -1 in `prepareToSave`. This ensures that the sound will be reallocated a slot in the list of saved sounds. Even if this sound has not changed another sound may have been added or removed, which could lead to this sound's old ID pointing to the wrong entry in the array.
> 
> Note that the list of saved sounds is rebuilt each save, so there's no need to deallocate the entry corresponding to the sound's old ID. See `recordImagesAndSounds` in `ProjectIO.as` for more details.

rschamp (approved):
These seem sensible to me. Thanks for the explanations.

jwzimmer:
Added `pr - good to merge` since we are including this change in today's deploy (it's on Staging already).

PR #1229, opened 2016-11-03
Enforce bounds in ListWatcher's setWidthHeight()

### LINKED ISSUE #1223: Improve handling adding a list watcher to the stage when the list is very large
The issue is that when a user has a project with a very large list being watched by a list watcher, the project may not be able to load (project page or editor). 

If the list takes a sufficiently long time to load, Flash kills the loading process as an "unresponsive script" well before it can complete, causing **error no. 1502** in the console.

This issue was reported via freshdesk no. 60427:

> Whenever I go onto my project it crashes my internet. It also won't load. Or it does load and keeps saying, "Plugin Stopped", it also has the reload button available. Can you help me? It just happens on my project called, [see ticket]. When I go on any other project it's fine. My dad said it's a website problem not an internet or computer problem. 

The project that isn't loading is 124673984.

I uploaded an older version of the project to their account and it does load (it has a shorter list).

@cwillisf was able to load the current version of the project (that isn't loading online) in the offline editor:

> it eventually finishes loading in the offline editor. It took about 90 seconds on my pretty-fast machine.

cwillisf:
The `ListWatcher` class creates a cell for each visible item as part of its resize handling (see `updateContents`). This can take a long time if the height of the list watcher and the length of its list are both very large. This is especially true at load-time, when a list watcher may not yet know what its true size should be.

This change enforces bounds in the `setWidthHeight()` method of the `ListWatcher` class. Bounding the size effectively bounds the number of list cells that will be created, eliminating long load-time delays.

This resolves #1223

### DIFF
--- a/src/watchers/ListWatcher.as
+++ b/src/watchers/ListWatcher.as
@@ -359,6 +359,14 @@ public class ListWatcher extends Sprite {
 	//------------------------------
 
 	public function setWidthHeight(w:int, h:int):void {
+		// Large (especially tall) list watchers can cause many, many list cells to be allocated, leading to delays.
+		// Bound the size so that updateContents() won't cause long delays for long lists.
+		var boundingObject:Object = this.parent || Scratch.app.stagePane || {width: 480, height: 360};
+		x = Math.max(0, Math.min(x, boundingObject.width - frame.minWidth));
+		y = Math.max(0, Math.min(y, boundingObject.height - frame.minHeight));
+		w = Math.max(frame.minWidth, Math.min(w, boundingObject.width - x));
+		h = Math.max(frame.minHeight, Math.min(h, boundingObject.height - y));
+
 		frame.setWidthHeight(w, h);
 		fixLayout();
 	}


thisandagain:
Looks good. Adding "needs-qa" label to this PR.\

/cc @jwzimmer

ericrosenbaum:
The test project loads fine in Chrome and Firefox on mac

jwzimmer:
**Testing - in progress**

Chrome version X.X... & Firefox X.X..., Staging, Mac

- [ ] Verify that project 124673984 doesn't load correctly on production
  - This may freeze or crash Flash or even your browser. I recommend not using your primary browser for this test.
- [ ] Download project 124673984 as an SB2 using the project recovery tool
  - If you're not familiar with this tool, ask me or @jwzimmer for help
- [ ] Verify that the SB2 loads correctly in the new online editor (v452 or above)
- [ ] Verify that the SB2 loads correctly in the new offline editor (v452 or above)

ericrosenbaum:
while testing, discovered a flash crash seemingly caused by scrolling a large list with a large item in it: #1235

PR #1246, opened 2016-12-05
Sprite menu self-referencing: resolve #1139

### LINKED ISSUE #1139: Which sprites/ options should be included in the block dropdowns, especially when they're moved to new contexts?
I noticed some confusing behaviour w.r.t. which Sprites show up in the dropdowns of a script when it is dragged to a new context (via the backpack, or via the sprite pane). I am not sure what the solution should be to making that behaviour clearer, especially given the scenarios brought up by @TheLogFather and @Airhogs777 around clones and programmatically-generated variables. However, given that the current behaviour is confusing/ inconsistent*, I think it would be worth considering if we can improve it as an enhancement. The other scenarios brought up regarding how users are actually using the values in the block dropdowns & their behaviour when they change contexts (via the backpack & other avenues) should be considered when thinking about what the behaviour should be, and the scope could be widened to include new features to support those use cases.

**Here is the confusing behaviour I noticed:**
A block with a sprite's name in it (e.g. "point towards [Sprite1]") will still have that Sprite's name in it when pulled out of the backpack for that sprite: What I mean is, if you have a block like "point towards [ ]", and you have Sprite1 and Sprite2, then when you use the block in a script for Sprite1, you would expect to see Sprite2 in the dropdown (so Sprite1 can point towards Sprite2). However, you can get the block to say "point towards [Sprite1]", which doesn't make sense, when it is in a script for Sprite1, if you use the backpack.

*I would consider this behaviour inconsistent since you can get Sprite1 as the value for a block in the script area for Sprite1 by dragging it into the backpack from Sprite2's scripting area, but you can't get Sprite1 as the value for a block in the script area for Sprite1 just by dragging that block from the block menu into the scripting area (you have to make the script in the area for Sprite2 and then use the backpack to change it's context to the script area for Sprite1).

Repro:
- Go to homepage
- Click on Create icon
- Create a second sprite >
- You can name them anything, but for the purposes of the repro I'll call them Sprite1 and Sprite2
- Drag the "point towards [ ]" block into the Script area for Sprite 2 (when Sprite 2 is active)
- Change the dropdown so the block says "point towards [Sprite1]"
- Drag the block into the backpack
- Click on Sprite1 (make Sprite1 active)
- Drag the block from the backpack into the scripting area
- Note that now you have a block saying that Sprite1 should "point towards [Sprite1]", which doesn't make sense
- If you use the block in a script, it will have no effect (which makes sense/ is better than other potential behaviour)

On production:
screen shot 2016-06-14 at 2 29 03 pm
On pre-flight:
screen shot 2016-06-14 at 2 22 16 pm

DanFessler:
Submitting this to resolve #1139.

In summary, these changes add the current sprite in all sprite drop-down menus.  While this may seem redundant, note that the sprite field is in reference to the base sprite and not it's clones.  Many projects on scratch already take advantage of this distinction for advanced clone management (either by using a variable reporter, or by back-pack kung-fu).  This doesn't change any functionality of scratch.  I've also added an instance reference, "myself", to the sprite attribute sensor - both for variable selection convenience, as well as enables for some programmatically-generated variable management on the instance as some people have already been doing.

Some possibilities:
Sprite clones can now detect if they are touching their own kind
Sprite clones can now point at the sprite root
you can now access properties of a sprite via a reporter

### DIFF
--- a/src/primitives/SensingPrims.as
+++ b/src/primitives/SensingPrims.as
@@ -92,10 +92,9 @@ public class SensingPrims {
 		}
 		if (!s.visible) return false;
 
-		;
 		var sBM:BitmapData = s.bitmap(true);
 		for each (var s2:ScratchSprite in app.stagePane.spritesAndClonesNamed(arg))
-			if (s2.visible && sBM.hitTest(s.bounds().topLeft, 1, s2.bitmap(true), s2.bounds().topLeft, 1))
+			if (s != s2 && s2.visible && sBM.hitTest(s.bounds().topLeft, 1, s2.bitmap(true), s2.bounds().topLeft, 1))
 				return true;
 
 		return false;
--- a/src/primitives/SensingPrims.as
+++ b/src/primitives/SensingPrims.as
@@ -250,7 +250,9 @@ public class SensingPrims {
 
 	private function primGetAttribute(b:Block):* {
 		var attribute:String = interp.arg(b, 0);
-		var obj:ScratchObj = app.stagePane.objNamed(String(interp.arg(b, 1)));
+		var objName:String = interp.arg(b, 1);
+		var obj:ScratchObj = app.stagePane.objNamed(objName);
+		if ('_myself_' == objName) obj = interp.activeThread.target;
 		if (!(obj is ScratchObj)) return 0;
 		if (obj is ScratchSprite) {
 			var s:ScratchSprite = ScratchSprite(obj);
--- a/src/uiwidgets/ScriptsPane.as
+++ b/src/uiwidgets/ScriptsPane.as
@@ -367,7 +367,7 @@ return true; // xxx disable this check for now; it was causing confusion at Scra
 
 	private function dropCompatible(droppedBlock:Block, target:DisplayObject):Boolean {
 		const menusThatAcceptReporters:Array = [
-			'broadcast', 'costume', 'backdrop', 'scene', 'sound',
+			'attribute', 'broadcast', 'costume', 'backdrop', 'scene', 'sound',
 			'spriteOnly', 'spriteOrMouse', 'location', 'spriteOrStage', 'touching'];
 		if (!droppedBlock.isReporter) return true; // dropping a command block
 		if (target is Block) {
--- a/src/uiwidgets/ScriptsPane.as
+++ b/src/uiwidgets/ScriptsPane.as
@@ -367,7 +367,7 @@ return true; // xxx disable this check for now; it was causing confusion at Scra
 
 	private function dropCompatible(droppedBlock:Block, target:DisplayObject):Boolean {
 		const menusThatAcceptReporters:Array = [
-			'attribute', 'broadcast', 'costume', 'backdrop', 'scene', 'sound',
+			'broadcast', 'costume', 'backdrop', 'scene', 'sound',
 			'spriteOnly', 'spriteOrMouse', 'location', 'spriteOrStage', 'touching'];
 		if (!droppedBlock.isReporter) return true; // dropping a command block
 		if (target is Block) {
--- a/src/scratch/BlockMenus.as
+++ b/src/scratch/BlockMenus.as
@@ -436,7 +436,7 @@ public class BlockMenus implements DragClient {
 		m.addLine();
 
 		for each (var sprite:ScratchSprite in app.stagePane.sprites()) {
-			if (includeSelf || sprite != app.viewedObj()) spriteNames.push(sprite.objName);
+			spriteNames.push(sprite.objName);
 		}
 		spriteNames.sort(Array.CASEINSENSITIVE);
 		for each (var spriteName:String in spriteNames) {
--- a/src/scratch/BlockMenus.as
+++ b/src/scratch/BlockMenus.as
@@ -92,7 +92,7 @@ public class BlockMenus implements DragClient {
 		if (menuName == 'spriteOnly') menuHandler.spriteMenu(evt, false, false, false, true, false);
 		if (menuName == 'spriteOrMouse') menuHandler.spriteMenu(evt, true, false, false, false, false);
 		if (menuName == 'location') menuHandler.spriteMenu(evt, true, false, false, false, true);
-		if (menuName == 'spriteOrStage') menuHandler.spriteMenu(evt, false, false, true, false, false);
+		if (menuName == 'spriteOrStage') menuHandler.spriteMenu(evt, false, false, true, true, false);
 		if (menuName == 'touching') menuHandler.spriteMenu(evt, true, true, false, false, false);
 		if (menuName == 'stageOrThis') menuHandler.stageOrThisSpriteMenu(evt);
 		if (menuName == 'stop') menuHandler.stopMenu(evt);
@@ -430,17 +430,13 @@ public class BlockMenus implements DragClient {
 		if (includeRandom) m.addItem(Translator.map('random position'), 'random position');
 		if (includeEdge) m.addItem(Translator.map('edge'), 'edge');
 		m.addLine();
-		if (includeStage) {
-			m.addItem(Translator.map('Stage'), 'Stage');
-			m.addLine();
-		}
-		if (includeSelf && !app.viewedObj().isStage) {
-			m.addItem(Translator.map('myself'), 'myself');
-			m.addLine();
-			spriteNames.push(app.viewedObj().objName);
-		}
+
+		if (includeStage) m.addItem(Translator.map('Stage'), 'Stage');
+		if (includeSelf && !app.viewedObj().isStage) m.addItem(Translator.map('myself'), 'myself');
+		m.addLine();
+
 		for each (var sprite:ScratchSprite in app.stagePane.sprites()) {
-			if (sprite != app.viewedObj()) spriteNames.push(sprite.objName);
+			if (includeSelf || sprite != app.viewedObj()) spriteNames.push(sprite.objName);
 		}
 		spriteNames.sort(Array.CASEINSENSITIVE);
 		for each (var spriteName:String in spriteNames) {


liam4:
@carljbowman you seemed to miss this when I [suggested](https://github.com/LLK/scratch-vm/issues/303#issuecomment-255869046) this for Scratch 3.0 :)

Obviously I'm :+1: for this, and I can't really see how it can be confusing in any way.. also, there's already a "create clone of (myself)" block, so an "(attribute) of (myself)" block makes sense!

tjvr:
Having both "current sprite" and "myself" in the same menu sounds confusing for beginners. I realise the distinction—one is always the original sprite, the other can be a clone—but it's quite subtle!

You say you've added "current sprite" to all menus—does that include e.g. "point towards _"?

That said; I want this feature!

ntlrsk:
This would add confusion for any use except for clones.

Airhogs777:
Maybe "myself" and "clones of myself"?

DanFessler:
So my implementation copies the pattern that was already in use for the "create clone of" block.  But I do understand the concern that it may be confusing terminology.  I think all confusion can be avoided if we change "myself" to "this clone".  Of course, if there is no clone, it would simply refer to the base sprite.

so sprite1's "[prop] of [sprite]" menu might look like this:
```
Stage
this clone
-------
sprite1
sprite2
sprite3
```

@tjvr I've added the current sprite to all sprite lists including the "point towards" - but it doesn't appear as the string "current sprite",  but rather just the sprites name in the list next to all the other sprites.  On the point towards block, it does nothing if it's referring to itself.  But if a clone uses it, it points to the base sprite.

tjvr:
> I think all confusion can be avoided if we change "myself" to "this clone"

That's certainly better!

Doesn't solve @ntlrsk's concern about including "Sprite1" in the menu for Sprite1, though.

src/primitives/SensingPrims.as
```diff
@@ -92,10 +92,9 @@ public class SensingPrims {
 		}
 		if (!s.visible) return false;
 
-		;
 		var sBM:BitmapData = s.bitmap(true);
 		for each (var s2:ScratchSprite in app.stagePane.spritesAndClonesNamed(arg))
-			if (s2.visible && sBM.hitTest(s.bounds().topLeft, 1, s2.bitmap(true), s2.bounds().topLeft, 1))
+			if (s != s2 && s2.visible && sBM.hitTest(s.bounds().topLeft, 1, s2.bitmap(true), s2.bounds().topLeft, 1))
```
tjvr, line 97: Probably best not to change this? It'll be difficult enough getting the menu changes merged :-)

src/uiwidgets/ScriptsPane.as
```diff
@@ -367,7 +367,7 @@ return true; // xxx disable this check for now; it was causing confusion at Scra
 
 	private function dropCompatible(droppedBlock:Block, target:DisplayObject):Boolean {
 		const menusThatAcceptReporters:Array = [
-			'broadcast', 'costume', 'backdrop', 'scene', 'sound',
+			'attribute', 'broadcast', 'costume', 'backdrop', 'scene', 'sound',
```
tjvr: This is also unrelated to #1139; probably best not to include it with this PR!

src/primitives/SensingPrims.as
DanFessler: This is necessary. Without this check, touching will always return true when referencing its own sprites/clones, even when none are touching, because it would test against its own instance.  If we're allowing touching blocks to self-reference, then this fix is needed as well.

src/uiwidgets/ScriptsPane.as
DanFessler: makes sense.  I can branch this as its own PR.

TheLogFather:
I think I've said much of what I wanted to say about certain aspects of this over at #1139.

I will say, though, that I kind of disagree with @ntlrsk about having the current sprite name in the "point towards" block's menu – I don't think "confusing" is quite the right word.

It's certainly a bit of a *surprise* (and that in itself can often be reason enough to avoid something in a UI, though that's more in terms of behaviour that results from doing/choosing something in that UI), but my experience of watching people work with computers (various interfaces and coding languages, including Scratch) is that real confusion can be categorised in two main ways: firstly, from something actually *behaving* in an apparently inexplicable way (i.e. after the person has chosen what they want to do, but it didn't, and continued not to, do as expected); secondly, from some expected/desired option/choice/button/behaviour/etc. *not* being available (i.e. when they don't have the ability to make the choice they want).

Conversely, however, there are many times when 'surprise' can actually be an *opportunity* rather than a problem, and I think this is such a time, and that Scratch should be looking to use such opportunities, since it is specifically one of its aims (i.e. to help teach, by providing a low barrier to experimenting with things to find out what they do).

TBH, at the moment the main confusion I've seen is from Scratch's *current* behaviour of *not* having such an item in those menus! (And that fits in with my second category above.)

As I said in #1139, I'm clearly in favour of adding the current sprite to many of these menus, including the "point towards" block, since in that case it is providing highly useful functionality – i.e. referencing the *original sprite* (which is not necessarily the same thing as "myself" or "this sprite/clone") – and I don't think having it in the menu is really confusing (just, on first seeing it, a surprise, that hopefully leads to questions and, even more hopefully, experimentation).

----

Speaking more generally, I think the Scratch Team should beware against thinking that causing low-level 'potential confusion' is reason enough by itself to avoid adding something. Not to mention that Scratch is already inherently confusing (and confused/inconsistent) in so many ways and on so many levels...

As a fairly trivial example: the meaning of a sprite in the "point towards" block is different to that in a "touching sprite" block – in the former it refers to only a single entity, the original sprite. In the latter it refers to multiple objects, i.e. the original sprite *and* all its clones.

Another example: the sensing behaviour of a hidden sprite – it can sense touching edge, colour, and mouse-pointer [which, by itself, regularly leads to desperate pleas for help, and/or bug reports, in the forums, let alone confusion and surprise!], yet it cannot sense touching any another sprite.

Yet another example – but this one much more subtle... the behaviour of different event hat blocks when the event reoccurs while its script is already running: the broadcast receivers and "when clicked" hat block scripts will terminate and restart when their event happens again, while the "key pressed" and "when timer/loudness" hat block scripts will ignore their events until they have finished. Why do they behave like this...? (You might recall the broken projects that resulted when the "key pressed" hat block was briefly changed to work like "when clicked"...)

----

But to get back to what I started to say in the previous section (before I got distracted by some of Scratch's current foibles)...

As is bound to happen with a programming language, many of the more subtle aspects of Scratch can only be found out through experience/explanation – in my first example above (the inconsistency between touching and point towards), the way it behaves is really the most useful, and I'm not suggesting it should be changed. I'm saying that it is something which must be learned by experience/explanation on reaching some level of developing projects with Scratch.

It's a mistake to think that Scratch will always avoid any 'potential confusion' (it's already *way* too late for that anyway!) Rather, as I said above, Scratch should be providing an easy way for Scratchers to try things out, to experiment, and to discover how things behave, and then to be able to develop reasoning why that should be so, and/or ask questions to find out answers. (And note that it does still require that those who develop Scratch should be looking to anticipate such questions, and providing sensible reasoning for its behaviour, even if it may not be obvious at first, so it should be clear that what I've said above doesn't mean you can just make it do whatever happens to take your fancy at the time...)

----

Huh, well, so much for my first sentence above...  :/

DanFessler:
> Doesn't solve @ntlrsk's concern about including "Sprite1" in the menu for Sprite1, though.

@tjvr I'm actually not too convinced that it would be confusing.  Building upon what @TheLogFather has said, For 'touching' and point to' blocks, I doubt people would even attempt to use them if they weren't using clones, let alone be confused by it, because the concept of "touching myself" or "pointing to myself" isn't logical otherwise.  And even if they did attempt to use them without clones, nothing bad happens, they just realize that what they're attempting won't result in anything.  Likewise, for the 'attribute of sprite' block, if they're not using clones, nothing bad happens either when using "myself" or "myspritename" - It will simply return the attribute of the sprite they expect it would.

tjvr:
> For 'touching' and 'point to' blocks, I doubt people would even attempt to use them if they weren't using clones, let alone be confused by it, because the concept of "touching myself" or "pointing to myself" isn't logical otherwise.

This actually convinced me that "myself" is more confusing than "this clone", IMHO.

TheLogFather:
So far, my comments above were directed specifically at adding the current sprite name as an item into various menus – I haven't said anything about adding "myself", or "this clone", etc. (I certainly don't think such a thing should be added into blocks like "point towards", "go to", "distance to" – only the current sprite name should be added.)

----

However... I would note that the way Scratch currently 'speaks' of clones in its blocks means that using "this clone" (instead of "myself") would also cause confusion. –Many Scratchers will consider the original sprite as *not* a clone, and would likely therefore not consider "this clone" could also apply to the original sprite.

I suspect "this sprite" would be better (though some might consider a clone as not quite being a 'real' sprite, so I also think that's not ideal – but maybe it's one of those times where we need to find a way, and actually make an effort, to speak such that it becomes clear that a clone of a sprite is just as much a sprite as the original...)

Also, we already have precedent for "this sprite" – in the "video motion/direction" block – and in that block it is a synonym for "myself" (i.e. it doesn't necessarily refer only to the original sprite, but to whatever object is running it, including Stage).

It could be argued, then, that the clone-creation block should be "create clone of [this sprite]" (rather than "...of [myself]") – in order to help clarify that a clone is truly a sprite.

However, I think that change could easily be misunderstood as always making a clone of the original sprite, rather than a clone of the clone – even though that menu would still also have the sprite's name itself as an item (and that _does_ make a clone of the original sprite, and _not_ of the clone).

----

OK, so the above discussion suggests to me that "myself" is preferable because there is no chance for confusion about what it refers to.

But it should only be added where it is appropriate (i.e. *not* in the "point towards", "go to", "distance to" blocks. And having "myself" in the "touching sprite" block's menu is questionable, since its only use would be as a form of self-visibility check (and, given how intensive touching tests can be in Scratch, I'd rather encourage use of a state variable [even though I've occasionally used this trick in the past to test sprite visibility]).

That leaves only the "[attribute] of [sprite]" block... and the only use I can see for adding "myself" in the right menu of that would be to allow more advanced Scratchers to use variable indirection within clones – and that obviously requires https://github.com/LLK/scratch-flash/pull/1250 (so without that I don't see any point to adding "myself").

----

In conclusion, then, as I made clear in my previous essay, I would argue there's no good reason to avoid having the current sprite name added into the menus in certain blocks where they're currently 'AWOL'.

I'd also argue that, if there is anywhere to add a self-referential menu item, then "myself" is preferable to both "this clone" and "this sprite".

However, I *don't* see any blocks where "myself" (or "this clone", or "this sprite") does need to be added, *unless* indirection also becomes supported in the "_ of _" sensor block by allowing expressions to be dropped into its left side. (And, while I'd certainly be glad to see that happen, I suspect it's not particularly high priority...)

And finally... if "myself" is preferable to "this sprite", doesn't it mean the video block should change...? :)

DanFessler:
@TheLogFather 

You make some excellent points about using "myself" vs "this sprite" vs "this clone".  Currently this PR makes no change on this issue (uses "myself"), and I'm fine leaving it that way if a better option is not clear.

> I certainly don't think such a thing should be added into blocks like "point towards", "go to", "distance to" – only the current sprite name should be added.

> ...it should only be added where it is appropriate (i.e. not in the "point towards", "go to", "distance to" blocks). And having "myself" in the "touching sprite" block's menu is questionable, since its only use would be as a form of self-visibility check

We are on the same page.  This PR does not add "myself" to either of those blocks, nor do I think it should.  The self-visibility check is something that hadn't crossed my mind, clever.  This PR actually prevents that behavior by making sure the two objects being tested are not the same instance.  This was necessary to provide the expected behavior of testing when clones of the same sprite are touching each other - as previously it would always return "true" regardless if two separate instances weren't touching because it would also test the instances against themselves.

> That leaves only the "[attribute] of [sprite]" block... and the only use I can see for adding "myself" in the right menu of that would be to allow more advanced Scratchers to use variable indirection within clones – and that obviously requires #1250 (so without that I don't see any point to adding "myself")

The one place this PR *did* add "myself" was to the "[attribute] of [sprite]" block.  Originally #1250 was part of this branch specifically to allow for indirection, but it was separate enough of an issue that I separated it into it's own PR.  Without it, the only compelling reason to add "myself" here is for convenience of selecting variables, and that's not much.  I'd be fine adding this change to #1250 instead if that's what is preferred.

DanFessler:
This has been sitting here for a while now without any conclusion.  Can we get an update on this topic?  I'd be more than happy to address any design considerations to make this pull request compatible with the goals of the scratch team

thisandagain:
Thanks for the contribution!

PR #1250, opened 2016-12-06
Accept reporters in attribute menu

DanFessler:
Very small change, but safe:
- allows reporters to be dropped on the attribute menu
- on the  "attribute of sprite/stage" block, If a reporter is dropped on the spriteOrStage menu, the attribute menu gives you the default list of sprite attributes (no variables, because the sprite is unknown)

The argument:
Scratch already allows reporters to be dropped on the sprite/stage field of an "attribute of sprite/stage" block.  If someone chose to do so, it's safe to assume that they're trying to access a sprite programmatically, not the stage, because the stage name is always the same.  The problem is once they do, the attribute menu assumes "stage" and they no longer have access to the list of sprite properties and variables (because the sprite is unknown until run-time), making the block useless in this scenario.  By presuming you want sprite attributes instead, we can at least provide the default list that is the same for all sprites.  If we also allow reporters to be dropped onto the menu, they could then specify a variable manually without the drop-down.

### DIFF
--- a/src/scratch/BlockMenus.as
+++ b/src/scratch/BlockMenus.as
@@ -217,14 +217,14 @@ public class BlockMenus implements DragClient {
 
 	private function attributeMenu(evt:MouseEvent):void {
 		// If all else fails, fall back to the stage menus (better than nothing?)
-		var obj:* = app.stagePane;
+		var obj:*;
 		if (block) {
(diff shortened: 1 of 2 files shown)


thisandagain:
Thanks for the contribution!

PR #1259, opened 2017-01-04
Remove all file extensions when loading a project

cwillisf:
When loading a project from a file, remove all instances of '.sb', '.sb2', or '.sbx' found at the end of the file name. Repeat until none remain.

Previous behavior: remove either '.sb' or '.sb2' exactly once. Since this didn't include '.sbx' there are now projects with titles like 'Silly.sbx.sbx.sbx' -- this was the motivation for making this check repeat.

Testing:
- Make sure the projects gets an appropriate title after loading various project files from local disk. For example, all of the following should result in a project title of "Example":
  - [ ] `Example.sb`
  - [ ] `Example.sb2`
  - [ ] `Example.sbx`
  - [ ] `Example.sbx.sbx.sbx`

This resolves LLK/scratchx#32

### DIFF
--- a/src/Scratch.as
+++ b/src/Scratch.as
@@ -620,8 +620,12 @@ public class Scratch extends Sprite {
 	}
 
 	public function setProjectName(s:String):void {
-		if (s.slice(-3) == '.sb') s = s.slice(0, -3);
-		if (s.slice(-4) == '.sb2') s = s.slice(0, -4);
+		for (;;) {
+			if (StringUtil.endsWith(s, '.sb')) s = s.slice(0, -3);
+			else if (StringUtil.endsWith(s, '.sb2')) s = s.slice(0, -4);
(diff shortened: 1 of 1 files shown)


TheLogFather:
In the (admittedly rather unlikely) event that a file is named just .sb/.sb2/.sbx ?
(–i.e. should the loop break out before it strips the name to nothing?)

cwillisf:
@TheLogFather in that case the title will be set to the empty string, which is a valid option. It's a little unusual, I agree, but both the editor and the web site can handle an empty title. If for some reason you want an empty title -- and ~kids~ people sometimes want strange things -- I think it's appropriate to preserve that through a save/load cycle.

Also: the save dialog will provide a default file name starting with `project` when you try to save a project with no title. This means that saving a file called `.sb2`, for example, will only happen if it's an intentional decision.

PR #1279, opened 2017-02-17
Warn about missing extension specs only on export

### LINKED ISSUE #1277: "Missing block specs" warning pop-up when you switch languages
@cwillisf noticed that if you switch languages in the editor on staging, you get a pop-up warning. 
> it's something I added so that we could translate the WeDo 2.0 blocks... I just didn't realize that code was called when you switch languages

screen shot 2017-02-17 at 3 25 52 pm

cwillisf:
The new `getExtensionSpecs` method can pop up a warning if one or more built-in extensions are not loaded. This helps avoid problems when exporting strings for translation, but we don't want this warning to pop up when users change their language setting. This change introduces a Boolean argument to control whether or not the warning should pop up.

This resolves #1277 which was introduced by #1274

### DIFF
--- a/src/translation/Translator.as
+++ b/src/translation/Translator.as
@@ -224,7 +224,7 @@ public class Translator {
 
 	private static function checkBlockTranslations():void {
 		for each (var entry:Array in Specs.commands) checkBlockSpec(entry[0]);
-		for each (var spec:String in Scratch.app.extensionManager.getExtensionSpecs()) checkBlockSpec(spec);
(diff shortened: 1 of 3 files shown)


src/translation/Translator.as
```diff
@@ -224,7 +224,7 @@ public class Translator {
 
 	private static function checkBlockTranslations():void {
```
cwillisf, line 225: This function is called when the user changes the editor's language setting.

rschamp (approved):

Issue #1276, opened 2017-02-15
Custom block boolean parameter can be sent 'bad' value after editing spec


TheLogFather:
As follows:
1) Create custom block `test (num)` where 'num' is a numerical input.
2) Create variable "var" and attach `set var to (num)` under definition.
3) Drag custom block caller into editing pane, and click on it.

Result: all fine so far: default value in caller is 1, so variable 'var' ends up as 1.

4) Edit custom block spec so it has a boolean instead of (num), i.e. to become `test <bool>`
5) Change the 'set var' block to `set var to <bool>`, so we see what 'bool' is.
6) Note caller now says `test < >`, so the boolean param is blank (which corresponds to false).
7) Click on caller.

Result: variable 'var' still gets set to 1 – but it should now be "false".
(Note that the boolean parameter 'bool' is supposed to be false, given the blank slot in the caller, but actually has the value 1. And that means in a boolean test it will equate to true – i.e. it'll end up doing completely the opposite of what you're expecting it to do.)

----

Examining the JSON shows what's happening:
`                [79.5, 93.5, [["call", "test %b", 1]]],`
The value of the param being passed into it is still one, from before the spec was changed (and we can make it send text too, if the first param was a string input, rather than numerical, before it was changed to boolean) – but, of course, that's not shown in the boolean param slot in the caller, since it only shows block expressions dropped into it.

----

It's easy to work around in the editor just by dropping something into the caller's boolean param slot (and then removing it) – but it had me really confused for a while in a project I was working on, 'cos I just couldn't understand how it could possibly be acting like the boolean param used in the custom block script was true when its corresponding caller slot was blank!

TheLogFather:
Made a bit of a fun project to demo how Scratch can use this to defy logic itself:
https://scratch.mit.edu/projects/147618731/
:)

(Interestingly, it's kinda related to what was just happening with the "not" block problem in #1284, which is why I managed to get to the bottom of that so quickly...)

milad-mehri:
Nice!

Issue #591, opened 2015-01-22
Lists - Duplication of list data


2jour:
From @TheLogFather,  "It looks like list info is getting placed in the JSON twice - appearing as part of the stage, as well as whatever else owns it...  This causes unreliable saves of long lists".

2jour:
TODO add more description @2jour

2jour:
Also issue with uploading projects with long list https://scratch.mit.edu/discuss/topic/170384/?page=1#post-1635847

thisandagain:
Adding `request` label to ensure we check this as a part of the implementation for 3.0

TheLogFather:
Now that the JSON limit has been reduced to 2.5MBytes, this issue has become rather more acute, and people are running into it more often. –Any chance of a fix being looked at sooner than 3.0 release?
(I guess one of the big issues is backward compatibility...?)

TheLogFather:
OK, just been having a look at this, and it seems to me that the main difference between variables and lists is that a variable is not the same as its watcher, whereas a list *is* the same as its watcher.

Now, variables are added to the JSON for sprite/stage here:
https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchObj.as#L592
And variable watchers for Stage are added here:
https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchStage.as#L839
And for sprites here:
https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchStage.as#L850

However, the lists are added to the JSON for sprite/stage here:
https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchObj.as#L593
But list watchers are also added for Stage and sprites in the same lines as above.

Since a the existence of a list already means the ListWatcher exists, isn't it simply a case of not pushing ListWatchers onto "children" in those two lines (839 and 850) – they've already been added in ScratchObj.as line 593.

I've experimented with simply removing the second instance of the lists from the JSON (which corresponds to not including ListWatchers as children), and it has worked for me so far...

TheLogFather:
To summarise previous comment (if TL;DR):

I think the *really simple* fix is to NOT include ListWatchers in lines 839 & 850 of ScratchObj.as.

TheLogFather:
(Oo-er... how to get rid of the multiple references above, where I messed up multiple times doing stuff with git/github in wrong branches, etc? –just for a change...)

PR #1284, opened 2017-02-24
Handle block arg text for args without text fields

### LINKED ISSUE #1283: Type Error 1009 - in Editor
**TL;DR:** I saw Type Error 1009 right after deploying when I dragged a script out of my backpack, played it by clicking it, then stopping it. @christanbalch also saw this error in the console when she ran into random projects not playing when she clicked green flag.

=== Details ===

Type Error Projects (all involving same scripts):
- https://scratch.mit.edu/projects/147019078/#editor
- https://scratch.mit.edu/projects/147020145/#editor
- https://scratch.mit.edu/projects/147021077/#editor

**Once you trigger the Type Error, changing the language doesn't work correctly, and the Scratch logo in the upper left doesn't take you to Scratch homepage. Neither does My Stuff.**

I think in my project I narrowed down the issue-causing script to this one:
screen shot 2017-02-24 at 12 17 29 pm

cwillisf:
Some blocks -- such as "set color to (color)" -- have a block argument which doesn't use a text field. The code to preserve user text was trying to collect the field text in these cases, resulting in an attempt to access a property of a null reference.

This change does a few things:
- Move the serialization of a `BlockArg`'s value into a method on   `BlockArg` instead of embedding this code into `BlockIO`'s `blockToArray()` function. This is (IMO) a better place for the code in general but also makes it easier to compare `getArgText()` with `setArgValue()`, which deals with similar concepts.
- Add a check to make sure that a block has a text field before trying to collect the text field's value.
- Make the conditions for going with the internal arg value (like `_mouse_`) instead of a potentially localized value (like "puntero del ratón" or "mouse-pointer") more similar to the conditions used by `setArgValue()`.
- Add more comments explaining the intricacies of this code. I have a feeling this stuff will be simpler in 3.0 but I wanted to record the reasoning behind this logic just in case.

This resolves #1283 
See also #1273 and #1278

/cc @jwzimmer

### DIFF
--- a/src/blocks/BlockArg.as
+++ b/src/blocks/BlockArg.as
@@ -36,13 +36,17 @@
 // arguments, it should set base to a BlockShape to support drag feedback.
 
 package blocks {
-	import flash.display.*;
-	import flash.events.*;
-	import flash.filters.BevelFilter;
-	import flash.text.*;
-	import scratch.BlockMenus;
-	import translation.Translator;
-	import util.Color;
+import flash.display.*;
+import flash.events.*;
+import flash.filters.BevelFilter;
+import flash.text.*;
+
+import scratch.BlockMenus;
+import scratch.ScratchObj;
+
+import translation.Translator;
+
+import util.Color;
 
 public class BlockArg extends Sprite {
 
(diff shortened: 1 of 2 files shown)


TheLogFather:
Wow, I see I missed all the fun yesterday with this 'green-flag glitch'!

Sorry – I kept meaning to check the previous args fix (for 'pick random'), since I wondered if there might be cases where it might not be quite right. Just never got around to it. :/

TheLogFather:
STOP PRESS!!

More coming...

cwillisf:
Thanks for the thorough testing, @TheLogFather!
I'm going to be out of the office for the rest of this week so we're definitely not going to rush this out. I'll dig into it as well; I'm sure that between us we can figure out which case(s) are still broken :(

TheLogFather:
OK, the problem is with boolean arguments for custom blocks.

More coming...

TheLogFather:
Actually, the problem is just with blank boolean slots. These get turned into "false", which is right by itself.

However, then using it with a a "not" operator *doesn't* switch it to true, 'cos <not "false"> is still false.

So, AFAICT, the only problem is with using that particular construction (i.e. a blank "not" operator block).

TheLogFather:
The other slight issue I think there is with this arg-to-string preservation is that it can, in certain cases, fairly significantly increase the size of the JSON – just because of the extra quotes around each item.

Considering how tight we can be already when there is list data in there (due to the list duplication issue #591), this could push some projects over the limit (if they have lots of numerical list items, which suddenly have two extra characters on each item).

cwillisf:
Hmm... those are some good points you make there, @TheLogFather. Sorry I missed the "not" block issue... I even tested Booleans! :/

I agree that it might be time to rethink the approach entirely. I'm hesitant to do it just for "pick random" but you do have a point about that being the main (maybe only) case where the exact details of the JSON round-trip matter. In principle I feel like saving & loading a project, duplicating a stack, or dragging a stack into and back out of your backpack should never change the contents of any block menus or fields but obviously that's harder than I expected...

Another option I've been thinking about: use the string if the argument is numeric and the field text doesn't match `String(Number(text))`. I wonder if there are projects which depend on our current handling of `1e1` or `e1.` or something... not that preserving such undefined behavior should be a priority.

TheLogFather:
Yeah, the blank "not" block is a little unusual, but it's fairly widely used as the 'quick' way to pass "true" into a custom block boolean slot (or anywhere else it's wanted).

I like that string comparison idea – it's more generic than targeting a specific block's args, but causes a change only where needed.

TheLogFather:
I've tried the following, and it looks like it does as you described above:

		for each (var arg:* in b.normalizedArgs()) {
			if (arg is Block) {
				result.push(blockToArray(arg));
			} else if (arg is BlockArg) {
				var argVal:* = BlockArg(arg).argValue;
				if (argVal is ScratchObj) {
					argVal = ScratchObj(argVal).objName;
				} else if (argVal is Number) { // special check for numerical values
					var argText:String = BlockArg(arg).getArgText();
					if (String(argVal)!=argText) argVal = argText; // if different then use text
				}
				result.push(argVal);
			}
		}

TheLogFather:
Actually, the recent issue above with the "not" block, along with the [demo project](https://scratch.mit.edu/projects/147618731/) I just made for #1276, cause me to wonder if applying "not" to the string "false" might actually be something Scratch should recognise, and return boolean true for that case...?
(I mean, Scratch recognises the string "false" as being equal to boolean false, so doesn't it makes sense for it to consider <not "false"> as boolean true?)

cwillisf:
Closing this in favor of #1287

Issue #1270, opened 2017-01-27
Scratch is converting "1.0" to "1" in numeric placeholders - breaks pick random block


griffpatch:
* Create a project with one script that reads (pick random -1.0) to 1.0).
* Save the project.
* Close and reload the project

Result is that the pick random has now been changed to -1 to 1.

This is bad because it results in a totally different results.

This did not used to be the case because I have projects where I use 1.0 for this very reason. This is also impacting Scratch 3 as it loads them without the decimal place too. This could be down to storing the arguments as numbers in the exported/saved JSON as this may trim the redundant decimal places?

CosmicWebServices:
I noticed this problem and I think I used a join block to fix it...

TheLogFather:
Was wondering when this would finally make its way into here... :)

@griffpatch: This happened a while back when the strange addition of an "epsilon" was removed from integer values: https://github.com/LLK/scratch-flash/pull/698

It's possible to work around it by 'hacking' a string value of "1.0" into the slot in the JSON (rather than a numeric value). –That doesn't get changed back to numeric 1, but remains as string "1.0".

cwillisf:
Here's my attempt at documenting the current behavior in Scratch 2.0 and what I believe to be the desired behavior. Please let me know if you think I've missed anything or have anything wrong, including if you think the desired behavior is different from what I've laid out below.

Assume that we're running a `random {} to {10}` block, and the value from the table's leftmost column is inserted as the first argument. I've used "int" and "float" here as shorthand for whether the editor will return only integer values or include fractional values.

| First argument | Current behavior | Desired behavior | Notes |
| - | - | - | - |
| 1 | int 1..10 | int 1..10 | |
| 1. | float 1..10 | float 1..10 | a |
| 1.0 | float 1..10 | float 1..10 | a |
| 1e-1 | float 1..10 | float 1..10 | b |
| 10e-1 | int 1..10 | float 1..10 | b |
| 10.e-1 | float 1..10 | float 1..10 | a, b |
| 1ee | int 0..10 | float 0..10 | b, c |
| 1.. | float 0..10 | float 0..10 | a, c |
| Reporter with value "1" | int 1..10 | int 1..10 | |
| Reporter with value "1.0" | int 1..10 | float 1..10 | d |
| Reporter with value "0.1" | float 0.1..10 | float 0.1..10 | |

(a) - Any string containing '.' is treated as non-integer.
(b) - IMO scientific notation should force float result for consistency.
[...]

PR #1287, opened 2017-03-06
Collect field text from numeric fields only

### LINKED ISSUE #1283: Type Error 1009 - in Editor

cwillisf:
This change preserves user-entered text for a block argument, but only in the specific case of a numeric block argument where the string representation of the numeric value doesn't match the field text. This means that if the user enters "42" then the serialized value will be a numeric 42, but if the user enters "42." then the serialized value will be that string exactly. Other argument types should not be affected by this change.

Resolves #1283 by taking a different approach to #1270
See also #1273 & #1284

Special thanks to @TheLogFather for lots of testing & discussion around this topic!

### DIFF
--- a/src/blocks/BlockIO.as
+++ b/src/blocks/BlockIO.as
@@ -72,7 +72,6 @@ public class BlockIO {
 		if (b.op == Specs.PROCEDURE_DEF)								// procedure definition
 			return [Specs.PROCEDURE_DEF, b.spec, b.parameterNames, b.defaultArgValues, b.warpProcFlag];
 		if (b.op == Specs.CALL) result = [Specs.CALL, b.spec];			// procedure call - arguments follow spec
-
 		// Note: arguments are always saved in normalized (i.e. left-to-right) order
 		for each (var arg:* in b.normalizedArgs()) {
 			if (arg is Block) {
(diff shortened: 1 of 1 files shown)


Kenny2github:
Is this going to be merged or not? It does seem like there’s a conflict - I wonder what the problem there is.

rschamp:
Thanks @Kenny2github. @cwillisf I don't remember where we ended up with this. We should either merge or close.

PR #1289, opened 2017-03-14
Fix #1288

### LINKED ISSUE #1288: Scratch's boolean operators applied to strings give 'unexpected' results (low-prio, but has very simple fix)
If a string is accidentally sent into a boolean custom block parameter, it's possible to get non-sensical results from Scratch (see http://scratch.mit.edu/projects/147618731 ).

This is related to https://github.com/LLK/scratch-flash/issues/1276 (and was also touched on in https://github.com/LLK/scratch-flash/pull/1284#issuecomment-282865943 ).

However, the fix is very simple...
Change the not operator in Primitives.as ([line 58](https://github.com/LLK/scratch-flash/blob/develop/src/primitives/Primitives.as#L58)) to return ````!interp.boolarg(b,0)````, instead of ````!interp.arg(b,0)````.

----

I have to admit that I'm a bit bemused that the "or" and "and" operator primitives seem to be working fine with both string "true" and string "false", as well as string "1" and string "0", even though they are also using "interp.arg"...

It seems as though ActionScript considers ("true" && "true") is true, and ("true" && "false") is false, yet it still considers that !"false" is false. (And similarly !"0" is false even though ("1" && "1") is true, and ("1" && "0") is false.) :/

EDIT: what I wrote above was wrong. See my next comment below. ("true" && "false") actually gives "true", but it's ("false" && "true") that gives "false". Similarly ("1" && "0") gives "1", while ("0" && "1") gives "0".

Since that behaviour seems a bit 'quirky', I'm wondering if it's a good idea to rely on it, and maybe the other boolean operators should also be changed to use interp.boolarg instead of interp.arg?

CosmicWebServices:
Thanks @thelogfather for the fix for #1288

### DIFF
--- a/src/primitives/Primitives.as
+++ b/src/primitives/Primitives.as
@@ -55,7 +55,7 @@ public class Primitives {
 		primTable[">"]				= function(b:*):* { return compare(interp.arg(b, 0), interp.arg(b, 1)) > 0 };
 		primTable["&"]				= function(b:*):* { return interp.arg(b, 0) && interp.arg(b, 1) };
 		primTable["|"]				= function(b:*):* { return interp.arg(b, 0) || interp.arg(b, 1) };
-		primTable["not"]			= function(b:*):* { return !interp.boolarg(b,0) };
+		primTable["not"]			= function(b:*):* { return !interp.boolarg(b, 0) };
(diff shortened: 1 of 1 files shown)


thisandagain:
Thanks for the contribution!

PR #1314, opened 2017-05-29
Magic variables

liam4:
This PR makes variables whose names start with `MAGIC_PROC_PREFIX` not appear visible in the block palette, unless currently editing a "Scrap Library block" (or a custom block whose spec starts with `MAGIC_PROC_PREFIX`).


liam4:
Haha. Oops.™

I knew I'd do this eventually..

So GitHub automatically selects LLK/scratch-flash as the base branch for pull requests. Inconvenient.

Issue #1041, opened 2016-03-11
LEGO WeDo 2.0 extension has no help


liam4:
![no-wedo2-help](https://cloud.githubusercontent.com/assets/9948030/13708414/e4016e6c-e784-11e5-807a-ef1deace2968.png)

Activating the dropdown menu for WeDo 2.0 and selecting About extension will link to this: https://cdn.scratch.mit.edu/scratchr2/static/help/en/ext/LEGO%20WeDo%202.0.html

..which doesn't exist.

ntlrsk:
Thanks for reporting. This is a known issue and the fix is going out soon. For now, please see: 
https://scratch.mit.edu/projects/editor/?tip_bar=ext2

cpseager:
BTW the offline 444.3 help link fails at a different position, it stops working here:

![captureoffline](https://cloud.githubusercontent.com/assets/7423187/13755006/0c3f5f7c-ea11-11e5-9a97-f88d4dac68ec.PNG)

cwillisf:
Fixed in Offline v444.4

PR #1335, opened 2017-08-08
Drum sample fix

christan1029:
Currently, when inside a repeat block, a drum block with "0 beats" re-starts the sound repeatedly with a tiny break in between, rather than not playing at all.

Referenced here: [#1041](https://github.com/LLK/scratch-blocks/issues/1041)

My change will create an exception when a user sets beats to 0. In this case, no sample will be played.


thisandagain:
This would break some existing projects. My inclination is to change this behavior in Scratch 3.0, but not introduce a breaking change in 2.0.

PR #1341, opened 2017-08-17
Work around Edge bug where charCode is 0

### LINKED ISSUE #1334: Key pressed not working for alphanumeric keys on Microsoft Edge
Hey everyone,

I ran into this issue with a project I'm building, and then found a forum thread on the issue:
https://scratch.mit.edu/discuss/topic/251930

It seems very serious. The <when key [... v] pressed?> blocks returns false, always, for any letter or number keys. Space bar and arrow keys seem to work fine. Key pressed event hats also don't trigger.

I shared this project to test this:
https://scratch.mit.edu/projects/170587322/
I can press all keys fine in Chrome, but only the arrows and the space bar in Edge.

Entering text in the editor, like for changing text values, works just fine. The issue only seems to affect projects.

Running Microsoft Edge 40.15063.0.0
Windows Version 10.0.15063 Build 15063

cwillisf:
### Resolves

This resolves #1334 

### Proposed Changes

When interpreting a Flash `KeyboardEvent` for the purposes of the `when {} key pressed` and `is key {} pressed?` blocks, read the `keyCode` property instead of the `charCode` property.

One difference between `keyCode` and `charCode` is that, for alphabetic keys, `keyCode` always returns a capital letter whereas the `charCode` value depends on the state of modifiers like the shift key and caps lock. Since Scratch relies on lower-case letters for `when {} key pressed` and `is key {} pressed?`, we convert to lower-case in all cases.

While I was in the area I adjusted some of the code so that there would be fewer "magic numbers" -- for example, I'm testing against `Keyboard.LEFT_ARROW` instead of 37 to recognize the left arrow key, and I'm using `toLowerCase()` instead of adding 32 to convert a letter to lower case. Also, I centralized the logic to collect a Scratch key identifier from a `KeyboardEvent` so that `keyDown` and `keyUp` no longer duplicate that code.

### Reason for Changes

Edge is delivering keyboard events with `charCode` = 0 even for basic letter or number keypresses. This seems to be a bug in Edge's implementation of Flash, but we can work around it by reading the `keyCode` property instead. The `keyCode` property seems reliable across all browsers.

### DIFF
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -37,6 +37,7 @@ import flash.media.*;
 import flash.net.*;
 import flash.system.System;
 import flash.text.TextField;
+import flash.ui.Keyboard;
 import flash.utils.*;
 
 import interpreter.*;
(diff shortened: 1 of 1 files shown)


src/scratch/ScratchRuntime.as
```diff
@@ -948,31 +949,37 @@ public class ScratchRuntime {
...
 
-	private function mapArrowKey(keyCode:int):int {
-		// map key codes for arrow keys to ASCII, other key codes to zero
-		if (keyCode == 37) return 28;
-		if (keyCode == 38) return 30;
-		if (keyCode == 39) return 29;
-		if (keyCode == 40) return 31;
+	// Get an ASCII value for the key pressed:
+	// - Latin letters will be normalized to lower-case
+	// - Arrows will be mapped to ASCII/Unicode delimiters (this is a traditional hack in Scratch 2.0; see mapArrowKey)
```
thisandagain, line 969: LOL @ "traditional hack"

thisandagain (approved):
This looks great. Thanks @cwillisf

Issue #1295, opened 2017-03-29
Support for other keyboard layouts (e.g. AZERTY) for 0-9 key detection


TheLogFather:
It's been brought to my attention that AZERTY keyboards don't have digits 0-9 in the top row (unless shift is also pressed), making it a bit awkward when Scratch doesn't cater for *any* of the characters that are produced by this top row of keys on AZERTY layout.

I was wondering about including some degree of support for this situation by adding the following line into functions ````keyDown```` and ````KeyUp```` in [ScratchRuntime.as](https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchRuntime.as#L949-L964), just before/after they ````mapArrowKey````:

````        if ((ch > 127) && (evt.keyCode >= 48) && (evt.keyCode <= 57)) ch = evt.keyCode;````

If the key's character code is  'out of range' then it checks if the key code belongs to the 'standard' top-row of keys (i.e. 0-9 on QWERTY). If so it then uses that as the character code (since the values match).

This would mean pressing that top row of keys acts like digits 0-9 in Scratch projects. I suspect it would also work for other layouts which don't have digits as the non-modified output for top-row keys.

Does anyone know of a reason why this could be a problem?

TheLogFather:
Hmmm, actually, now that I try that, I see it doesn't work as well as I thought, since the character codes produced by several of those unmodified keys are (obviously) <128.

E.g:
1  -->  &
3  -->  "
4  -->  '
5  -->  (
8  -->  !

Since all of those are 'unsupported' keys, only available through 'hacked' blocks, those cases could possibly (maybe, vaguely...) be 'ignored' – but has anyone got any other (better) bright ideas...?

TheLogFather:
OK, so I don't really know what the 'best' thing is to do with this, but perhaps it might be nice to increase the ````keyIsDown```` array to hold 256 bools instead of 127...?
That would allow 'hacked' blocks such as <key [ç] pressed?> to work, so at least the top row on AZERTY no longer has totally unusable keys in Scratch (i.e. é, §, è, ç, à).

PR #1345, opened 2017-08-21
Use `keyCode` instead of ASCII for `primKeyPressed`

cwillisf:
The `keyIsDown` array now tracks keys by their normalized `keyCode` rather than ASCII value. This means that in situations where we don't get a good `charCode` (see #1334) we can still track a key without worrying about mapping its `keyCode` to an ASCII value. That mapping is straightforward for some keys (letters, for example) but not for others (comma, for example).

This change is the result of a problem that @tarmelop noticed with #1341 where certain keys would work for the `when {any} key pressed` hat block but would not work for the `is {any} key pressed?` reporter. These are keys like the comma, where the `keyCode` value doesn't match the character's ASCII value -- this was the key oversight with #1341. Also, the `keyCode` value for some such keys is larger than 127, so that restriction has been removed from the `keyIsDown` handling.

## Testing

This code is currently available for testing on my dev port (cwillisf).

Put code like this on the default cat:
image

If you press and hold the key indicated, then the cat should:
- Repeatedly move its legs (costume change), and
- Rotate once, delay (key repeat delay), then repeatedly spin in place.

Overall this makes it look like the cat starts running in place, then gets traction and starts running in circles. I believe "cartoon run" is the technical term for this behavior. 🤓

Note that Edge sometimes caches the SWF even through a "hard" reload. I recommend visiting a page other than the editor page then clearing the cache before testing. Clear the cache by opening the "..." menu, picking "Settings", "Choose what to clear", and then "Clear" with at least "Cached data and files" checked.

Test with at least the following keys:
- The 'space' key
  - [ ] cartoon run works
- At least one arrow key
  - [ ] cartoon run works
- At least one letter key
  - [ ] cartoon run works
  - [ ] modifier states don't matter (shift, caps lock)
- At least one number key
  - [ ] regular number key causes cartoon run
  - [ ] numeric keypad number key causes cartoon run (check numlock)
- 'any' key
  - [ ] space key causes cartoon run
  - [ ] arrow key causes cartoon run
  - [ ] letter key causes cartoon run
  - [ ] regular number key causes cartoon run
  - [ ] numeric keypad number key causes cartoon run
  - [ ] comma key causes cartoon run

### DIFF
--- a/src/primitives/SensingPrims.as
+++ b/src/primitives/SensingPrims.as
@@ -25,6 +25,7 @@
 package primitives {
 	import flash.display.*;
 	import flash.geom.*;
+	import flash.ui.Keyboard;
 	import flash.utils.Dictionary;
 	import blocks.Block;
 	import interpreter.*;
@@ -229,14 +230,16 @@ public class SensingPrims {
 			}
 			return false;
 		}
-		var ch:int = key.charCodeAt(0);
-		if (ch > 127) return false;
-		if (key == 'left arrow') ch = 28;
-		if (key == 'right arrow') ch = 29;
-		if (key == 'up arrow') ch = 30;
-		if (key == 'down arrow') ch = 31;
-		if (key == 'space') ch = 32;
-		return app.runtime.keyIsDown[ch];
+		var ch:int;
+		switch (key) {
+			case 'left arrow': ch = Keyboard.LEFT; break;
+			case 'right arrow': ch = Keyboard.RIGHT; break;
+			case 'up arrow': ch = Keyboard.UP; break;
+			case 'down arrow': ch = Keyboard.DOWN; break;
+			case 'space': ch = Keyboard.SPACE; break;
+			default: ch = key.toUpperCase().charCodeAt(0); break;
+		}
+		return app.runtime.keyIsDown[ch] || false;
 	}
 
 	private function primDistanceTo(b:Block):Number {
--- a/src/scratch/ScratchRuntime.as
+++ b/src/scratch/ScratchRuntime.as
@@ -26,6 +26,8 @@ import assets.Resources;
 import blocks.Block;
 import blocks.BlockArg;
 
+import com.adobe.utils.StringUtil;
+
 import extensions.ExtensionManager;
 
 import flash.display.*;
@@ -68,8 +70,7 @@ public class ScratchRuntime {
 	public var app:Scratch;
 	public var interp:Interpreter;
 	public var motionDetector:VideoMotionPrims;
-	public var keyIsDown:Array = new Array(128); // records key up/down state
-	public var shiftIsDown:Boolean;
+	public var keyIsDown:Array = []; // sparse array recording key up/down state
 	public var lastAnswer:String = '';
 	public var cloneCount:int;
 	public var edgeTriggersEnabled:Boolean = false; // initially false, becomes true when project first run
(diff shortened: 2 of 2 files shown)


thisandagain:
Hmm. Good catch from @tarmelop. Thanks for the fix @cwillisf. This looks good.

griffpatch:
Out of interest does this break projects that make use of extended scratch
runtime features (hacked blocks).

I think this project used them so detect key presses?

https://scratch.mit.edu/projects/126517840/

cwillisf:
@griffpatch I'm afraid it would break that method, yeah. Specifically:
- Number keys still work fine
- Letter keys are double-detected: for example, pressing the 'A' key is detected by that project as 'A' and 'a'. (Because of the `toUpperCase` on [this line of code](https://github.com/LLK/scratch-flash/pull/1345/files#diff-c4e46451a2d78d494c2e6cf29da75248R240))
- Most other keys -- punctuation, for example -- are no longer detected at all, or are mis-detected. Examples:
  - The comma key is not detected
  - The home key is detected as `$`

Nearly all of the problems can be fixed by changing some of the code on the "ASCII Keys" sprite, but it's a bit of a pain since you'll have to handle mapping from the Flash [`Keyboard` constants](http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/ui/Keyboard.html#summaryTableConstant) to ASCII values for things like punctuation. You'd also need to watch for `Keyboard.SHIFT` (16) to determine whether or not to capitalize letters. I can't think of any way to check for caps-lock state, though... that's the only thing I can think of that this would break permanently.

Kenny2github:
Oh noo! One of my projects relies completely on hacked blocks - that’s broken now...

cwillisf:
@Kenny2github yeah, sorry about that. The good news is that the changes I described above should work for any project using "hacked blocks" to detect keypresses like this...

TheLogFather:
Any testing done for this with different keyboard layouts? (And how it might interact with #1295...?)

From my own testing on my MacBook's keyboard, it seems to totally ignore when I switch layout, so it's always using the actual layout of the laptop's keyboard – would be interested to see what it does with a truly AZERTY-layout keyboard attached (which I don't have to hand, unfortunately...)

Also, I see that it ignores any changes made by the shift key (so it cannot detect exclamation mark, for example, which is shift one on my keyboard). –I wonder if that means it'll be impossible to detect numerical keys on the top row of AZERTY keyboards? (Since it currently needs the shift key held down to access them – though keypad numbers should probably work ok...?)

It'd be useful to test this more extensively on an actual AZERTY keyboard before it's rolled out, I think...

jwzimmer:
As a result of discussing testing with @cwillisf: https://github.com/LLK/scratch-ops/issues/303.

Kenny2github:
@jwzimmer that's a private repo, I hope you know.

cwillisf:
@Kenny2github, thanks for the concern! In case you or anyone else is curious, the [scratch-ops](/LLK/scratch-ops) repository is where we hold internal discussions regarding operations ("ops") on the Scratch team, including occasional discussions about policy and support decisions. In this case, @jwzimmer was just making a note that we should discuss and formalize our support and testing policies regarding keyboard layouts.

Kenny2github:
@cwillisf ah, I see. Good to know!

cwillisf:
I tested with an AZERTY layout (Belgian, specifically) and found that it works reasonably well. Specifically:
- Letters are detected correctly according to the AZERTY layout. For example, when I press "A" on the AZERTY layout it is detected as A -- not as Q, which is the key in that position on a QWERTY keyboard.
- Pressing accented letter keys trigger the "any" key blocks, but are otherwise not detected. For example, pressing the Ù key is not detected as a U.
- Number keys are detected as numbers whether or not the shift key is pressed. For example, pressing the É key registers as a 2, since on this Belgian layout you would type a 2 by pressing shift and É together. This might be unexpected at first, but it seems like a good thing with respect to #1295.

Unfortunately this change does make it impossible to accurately implement some of the hacked-blocks projects. For example, let's say you're implementing the keyboard handling from the project mentioned above: https://scratch.mit.edu/projects/126517840/

Let's say that the user presses the key which Scratch, as of this change, now recognizes as 2. Let's further assume that you're tracking the state of the shift key -- now recognized as keycode 16. Unfortunately there are four possible results:

Keyboard layout | Shift not held | Shift is held
--- | --- | ---
en-US QWERTY | 2 | @
fr-BE AZERTY | É | 2

I'm not aware of any way for your Scratch project to detect the keyboard layout, so unfortunately you won't be able to decide which row to use. Basically, we've moved #1295 so that it only affects hacked blocks. Unfortunately I don't think we can solve #1334 without breaking the hacked blocks case. Sorry :(

griffpatch:
You could solve the keyboard input problem with a new reporter block that
returns a single character string or empty string that is a buffered stream
of keyboard input characters. But you would probably need a buffer reset
block too which would make it too complex I expect :(. Alternative a block
that returns a string of all input since the last time it was used (with a
max capacity of some given character length).

TheLogFather:
Just wondering if there's a way to use TextEvent instead to get the appropriate character...?
http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/TextEvent.html

PiotrGrochowski:
The shift issue referenced above refers to the fact that shift-click no longer erases in my EE remake. https://scratch.mit.edu/projects/168423636

PR #1360, opened 2017-09-25
Creating Scatch as General Programming porpose 

LaksahanWijerathne:
This Is Only And idea please don't confuse with these comment.

Needed advantages:
* User can be create new blocks and re-use  & upload them to online server to share with other
* program can be cover all areas like advance programming languages cover 
* into single code line blocks and combined block can be use as single line block, the properties window show that changeable properties  
* Preview window, Properties window, working block code window, text coding windows can be use as whenever user want them.
* User interface should be very attractive and user friendly
*


thisandagain:
Scratch 2.0 is now in maintenance mode while the team focuses efforts on [Scratch 3.0](https://scratch.mit.edu/developers). While critical issues will be addressed please note that any feature requests or minor issues will not be reviewed until the next major release. If you have requests or suggestions for the next version of Scratch please [file an issue here](https://github.com/llk/scratch-gui/issues) or [start a discussion here](https://scratch.mit.edu/discuss/1/).

PR #1365, opened 2017-10-27
Pseudo-revert keypress changes

### LINKED ISSUE #1361: Letter key presses not detected and quick presses in succession not detected - Edge + key press detection
Potential new issues being reported with Edge + keyboard detection, reported via https://scratch.mit.edu/discuss/topic/251930/ (the more recent pages/ posts):

```
I've run the keyboard test project below and it seems there are two problems when using Edge:
1. The bug where letter keys do not trigger as being pressed.
2. Another bug where pressing and holding a key and then immediately pressing a different key causes the second key to not be recognized.  Here's the repro case:
  - Press and hold the left arrow button for 10 seconds.  The left arrow button is shown as being pressed
  - Release the left arrow button and immediately press the right arrow button.
EXPECTED:  The left arrow shows as being not-pressed and the right arrow shows as being pressed.
OBSERVED: The left arrow continues to show being pressed for another 2-4 seconds after its key-up, and the right arrow doesn't show being pressed at all.
Can anyone else confirm this result?
```

@cwillisf said this could be a bug / is worth trying to repro, "probably introduced in v457. If the latest Edge is fixed then maybe we should consider removing the keyboard change I made in v457..."

cwillisf:
This resolves #1361 to the extent that we can. Note that #1361 is `prio-critical` so this could be a candidate for a hotfix, or maybe just late inclusion into the current freeze? CC @jwzimmer

It may also be interesting to compare this code to the code we had before #1341. You can simulate this by just looking at the last commit: 807e2b0e3cc133f3c5d449a7fc82437678313cde

Edge has fixed their `charCode` bug, so we don't need to work around it any more; this change starts off by reverting #1341 and #1345. However, I felt that some of the changes I made for the workaround were worth keeping. This code should have the same observable behavior as the pre-Edge-fix code, but logic related to keyboard events and key codes & names is centralized and better documented. Hopefully this will make this a more useful reference as we run into edge cases and quirks in Scratch 3.0...

Note that there still seem to be bugs in Edge: sometimes keys can become "stuck" for a little while, and keyboard response seems generally a bit slower than in other browsers. I've verified that these issues affect other Flash apps so I don't think there's anything we can do about it.

Testing:
In several different browsers, including Edge:
- Load & run my "keyscan" test project (182455392 from production or rename [this ZIP](https://github.com/LLK/scratch-flash/files/1423352/keyscan.zip) to end in SB2)
  - While a key is held it should appear in the "key pressed?" list, and the last few keys pressed should be listed in the "when key" list.
- Verify that several different types of key work as expected (here's an [ASCII table](https://en.wikipedia.org/wiki/ASCII#Printable_characters) in case it helps):
  - [ ] letters
  - [ ] numbers
  - [ ] punctuation & other symbols (like `[` or `/`)
  - [ ] space (32) & tab (9)
- Load & run "Tale of the Fiery Dragon v1.5c" by Griffpatch (126517840 from production)
  - Click past the initial screen of text
  - [ ] Verify that the text prompt works as expected

Note: it might help to compare behavior with scratchx.org, which is currently using a version of the editor from before #1341.

### DIFF
--- a/src/primitives/SensingPrims.as
+++ b/src/primitives/SensingPrims.as
@@ -25,7 +25,6 @@
 package primitives {
 	import flash.display.*;
 	import flash.geom.*;
-	import flash.ui.Keyboard;
 	import flash.utils.Dictionary;
 	import blocks.Block;
 	import interpreter.*;
@@ -230,16 +229,14 @@ public class SensingPrims {
 			}
 			return false;
 		}
-		var ch:int;
-		switch (key) {
-			case 'left arrow': ch = Keyboard.LEFT; break;
-			case 'right arrow': ch = Keyboard.RIGHT; break;
-			case 'up arrow': ch = Keyboard.UP; break;
-			case 'down arrow': ch = Keyboard.DOWN; break;
-			case 'space': ch = Keyboard.SPACE; break;
-			default: ch = key.toUpperCase().charCodeAt(0); break;
-		}
-		return app.runtime.keyIsDown[ch] || false;
+		var ch:int = key.charCodeAt(0);
+		if (ch > 127) return false;
+		if (key == 'left arrow') ch = 28;
+		if (key == 'right arrow') ch = 29;
+		if (key == 'up arrow') ch = 30;
+		if (key == 'down arrow') ch = 31;
+		if (key == 'space') ch = 32;
+		return app.runtime.keyIsDown[ch];
 	}
 
 	private function primDistanceTo(b:Block):Number {
--- a/src/primitives/SensingPrims.as
+++ b/src/primitives/SensingPrims.as
@@ -229,14 +229,8 @@ public class SensingPrims {
 			}
 			return false;
 		}
-		var ch:int = key.charCodeAt(0);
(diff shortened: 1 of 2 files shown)


thisandagain (approved):
Thanks @cwillisf. This is testing well for me locally, but I'm running inside a VM. I think it's wise to land this and deploy it to staging so we can do more thorough testing.

cwillisf:
New v459 SWFs containing this change are now available in the `develop` branch of `scratchr2`.
@thisandagain @jwzimmer

mewtaylor:
On Chrome 62, Mac OSX 10.12 on staging:
* Keyscan project loads, shows numbers that I would expect for keys pressed
* Tale of the fiery dragon works as expected

@cwillisf looks alright to me!

rschamp:
LGTM @cwillisf. I had issues with my caps lock key which is remapped to ctrl but I think that's an edge case.

cwillisf:
@rschamp: do you have the same issues on scratchx.org? Also... wat?

PR #1374, opened 2017-12-19
Telemetry system related updates

sammoh94:
Minor changes needed for the telemetry system


src/translation/TranslatableStrings.as
```diff
@@ -128,6 +128,7 @@ public class TranslatableStrings {
 		addAll(VariableSettings.strings());
 		addAll(Watcher.strings());
 		addAll(CameraDialog.strings());
+		addAll(PrivacyDialog.strings());
```
cwillisf: It looks like you might have forgotten to add `PrivacyDialog.as` to this branch. That said, if possible I think it would be a good idea to add it in `scratch-flash-offline` instead since it won't be used by the online editor or by ScratchX.

src/Scratch.as
```diff
@@ -135,6 +136,9 @@ public class Scratch extends Sprite {
 
 	public var logger:Log = new Log(16);
 
+	// Offline Queue
+	private var offlineSystem:OfflineSystem = OfflineSystem.getInstance();
+
```
cwillisf: I think it would be best to move this line into the `scratch-flash-offline` repository, along with `OfflineSystem.as`.

src/Scratch.as
```diff
@@ -1136,11 +1140,13 @@ public class Scratch extends Sprite {
 		function clearProject():void {
 			startNewProject('', '');
 			setProjectName('Untitled');
+			var clientID:String = offlineSystem.getClientId();
+			var appOpenPacket:Packet = new Packet(clientID, "Untitled", 'project', 'create', offlineSystem.getCurrentLanguage(), null);
+			offlineSystem.enqueuePacket(appOpenPacket);
```
cwillisf: As far as I can tell this is the only tricky part of moving the `offlinesystem` variable into `ScratchOffline` in the `scratch-flash-offline` repository. I would recommend replacing these three lines with a call to, say, `onNewProject()`. In the `Scratch` class, you can just make an empty protected method. Then, in `ScratchOffline`, make an override for `onNewProject()` which contains these three lines.

That would keep the offline telemetry system contained within the `scratch-flash-offline` repository, and the only changes in `scratch-flash` would be supporting-but-generic changes.

Let me know if there's a reason that this won't work; maybe I can help come up with a solution.

cwillisf (changes requested):
The functionality looks good but I'd like to move the stuff that's specifically offline-only into the `scratch-flash-offline` repository if possible.

src/translation/TranslatableStrings.as
sammoh94: @cwillisf PrivacyDialog.as should be included in `scratch-flash-offline`. https://github.com/LLK/scratch-flash-offline/blob/offline-editor-telemetry/src/PrivacyDialog.as

cwillisf: Ah, I see. In that case the call to `PrivacyDialog.strings()` will need to be in `scratch-flash-offline` as well: as it is, this will cause a build error for the online editor and ScratchX.

cwillisf (approved):

Kenny2github:
What even was this...?

towerofnix:
Judging by the offline code in those dead commits (a9dca1f92021de7c40bf365fa883dfd2e03877dc in particular), they added a way to track when people create new projects. (On the offline editor, it seems?)

There's [a change](https://github.com/LLK/scratch-flash/pull/1374/commits/38b4ac1359b9fa0993f2e7754cd758e86a2c929d) towards adding a privacy dialog. That was moved to scratch-flash-online, but still shows they'll probably have a popup of some sort, maybe asking the user for permission to track things like making a new project, or maybe just alerting the user that they do track that.

There's also [a change](https://github.com/LLK/scratch-flash/pull/1374/files#diff-49f5949cc990fddced317753606590f1R216) to add some padding to the dialog box element. I wonder how that's going to be used?

cwillisf:
This is all related to a system being developed by @sammoh94 which will let people who use the offline editor choose to send us information about how they use the editor. It has some fancy features to deal with situations where the Internet is not available, but the basic idea is we'd like to know how we can make offline Scratch better in the future (including Scratch 3.0). We call it the offline editor [telemetry](https://en.wikipedia.org/wiki/Telemetry#Software) system.

In the online editor we can watch our servers to see when certain things need to be improved, but we can't really do that with the offline editor since, you know, it's offline. This system is designed to help us with that limitation while still respecting the privacy of people using Scratch.

This particular PR doesn't do much on its own: it's just a few changes to help the offline editor telemetry system do its job. The telemetry system in general is still in progress, but will probably be ready very soon.

PR #1398, opened 2018-06-21
Fix warp thread bug caused by Flash 30 perf "fix"

### LINKED ISSUE #1397: Since PR #1396, the "timer" block no longer gets updated correctly during non-refresh loops
See https://github.com/LLK/scratch-flash/pull/1396#issuecomment-396726678 for more discussion...

cwillisf:
The recent changes made in #1396, intended to fix performance problems on Flash version 30, also introduced a timing problem in Scratch threads running in warp mode. Timekeeping was wrong in such cases, leading to incorrect timer block results (see #1397) and causing minutes-long or even permanent hangs in some projects.

This change resolves #1397 by correcting the timekeeping for warp-mode threads, removes the potential for hangs, and also simplifies the thread iteration count estimation math.

I suggest testing this code against at least these three projects:
- 153778200 Natura's Lake / Natura's Adventure
  - Try to jump out of the water on the first screen with an enemy fish.
  - Pass: noticeable hitches.
  - Fail: freeze for several seconds / minutes / forever.
- 10125047 Lag test
  - Pass: project runs, but maybe feels a little slow.
  - Fail: project hangs or runs extremely slowly.
- 227384326 Dungeon Glory
  - Pass: you can wander around the dungeon at an acceptable framerate.
  - Fail: project hangs or runs extremely slowly.

CC @TheLogFather

### DIFF
--- a/src/interpreter/Interpreter.as
+++ b/src/interpreter/Interpreter.as
@@ -221,22 +221,12 @@ public class Interpreter {
 	}
 
 	private const workTimeCheckIntervalFactor:Number = 1/3.0;
-	private const maxIterationCountSamples: uint = 10;
-	private var iterationCountSamples: Vector.<uint> = new <uint>[500]; // initial guess
+	private var iterationCountEstimate: Number = 1; // start low to avoid appearance of hang
 
 	private function addIterationCountSample(sample:uint):void {
-		iterationCountSamples.push(sample);
-		while (iterationCountSamples.length > maxIterationCountSamples) {
-			iterationCountSamples.shift();
-		}
-	}
-
-	private function getAverageIterationCount():Number {
-		var total:uint = 0;
-		for each (var sample:uint in iterationCountSamples) {
-			total += sample;
-		}
-		return Number(total) / iterationCountSamples.length;
+		// Exponential moving average: simulate N=10 with a=2/(N+1)=2/11
+		const alpha:Number = 2.0/11.0;
+		iterationCountEstimate += alpha * (sample - iterationCountEstimate);
 	}
 
 	public function stepThreads():void {
@@ -244,9 +234,8 @@ public class Interpreter {
 		doRedraw = false;
 		startTime = currentMSecs = CachedTimer.getFreshTimer();
 		if (threads.length == 0) return;
-		var currentEstimate:Number = getAverageIterationCount();
 		var iterationCount:uint = 0;
-		var checkInterval:uint = Math.round(workTimeCheckIntervalFactor * currentEstimate);
+		var checkInterval:uint = Math.round(workTimeCheckIntervalFactor * iterationCountEstimate);
 		var checkCount:uint = 0;
 		while ((currentMSecs - startTime) < workTime) {
 			if (warpThread && (warpThread.block == null)) clearWarpBlock();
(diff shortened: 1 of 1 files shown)


thisandagain (approved):
153778200, 10125047, and 227384326 all run quite well with this change. This logic itself is nice (actually a simplification over the last PR). Thanks for putting this together.

src/interpreter/Interpreter.as
```diff
@@ -296,15 +285,13 @@ public class Interpreter {
 			}
 		}
 		yield = false;
-		var warpStartTimer:int = CachedTimer.getCachedTimer();
 		while (true) {
-			if (activeThread == warpThread) currentMSecs = warpStartTimer;
+			if (activeThread == warpThread) currentMSecs = CachedTimer.getFreshTimer();
```
TheLogFather, line 289: Wondering if using getFreshTimer here for every 'top-level' block (i.e. every pass through this loop) will cause a significant slowdown during non-refresh under the new Flash 30? (I haven't upgraded Flash yet, so I don't know just how much of a slowdown Adobe's hobbling of getTimer caused...)

TheLogFather:
Any chance this could get bumped? –There have been a number of cases of 'random' crashes and issues reported in the forums due to this. Here are a few I noticed, for example:
https://scratch.mit.edu/discuss/topic/307677/
https://scratch.mit.edu/discuss/topic/312487/
https://scratch.mit.edu/discuss/topic/305171/
https://scratch.mit.edu/discuss/topic/312269/
https://scratch.mit.edu/discuss/topic/313729/
https://scratch.mit.edu/discuss/topic/313993/
(And there are numerous occasions I've had to explain to someone why many of my speedtest projects are giving zero for their timings, as well as why a few of my projects crash because they try to use the timer to know when to leave non-refresh.)

TheLogFather (commented):
I finally got around to downloading latest Flash and looking to see how much of a performance hit Abode put onto getTimer... and, wow, yes... a lot!

I've run a simple test of a non-refresh loop using v460 (https://github.com/LLK/scratch-flash/commit/35b6a84bfd508c17105d6bf68cd0962a5c8c94a0), v461 (https://github.com/LLK/scratch-flash/commit/76af421d3eff6ad3984de444738b1eae3148aa64) and v461.1 (https://github.com/LLK/scratch-flash/commit/74ae0cbc0427d6acc0508e41a73a324f84ebc8c4) under Flash 29 vs 31.

As expected, all three are similar running within Flash 29. (Interesting that v461 is ~10% faster – I guess not actually calling getTimer has a measurable effect even before it was 'hobbled' by Adobe...)

Unfortunately, with Flash 31, I'm still finding performance under v461.1 is just as badly affected as it is for v460. Of course, v461 is fine (since it doesn't use getTimer in the non-refresh loop).

If my understanding is correct, v461.1 isn't meant to be calling getTimer as much as v460 (since it does it only once per stack block, rather than every block?), so it seems a bit odd that it's still just as bad. (Maybe Adobe has slowed everything down based more on count of calls to getTimer within a certain time...?)

Have the Scratch Devs already found the above? (I.e. that applying this warp thread bug fix causes regression back to major slowdown.) Is that why this fix hasn't been deployed?

----

OK, so above is pretty problematic – but here's something positive...

I've been using the "days since 2000" block, and that doesn't cause such a major slowdown.

That suggests to me that it'd be possible to replace getTimer() with a time difference worked out through "new Date()".

Under Flash 29, using "days since 2000" is definitely slower than using the "timer" block (about twice), but it seems like it's the best option available, and it may be possible to do some caching of its value over stack blocks (i.e. only call it after a certain number of stack blocks, rather than every one, perhaps with certain types of stack block counting more, since they can potentially take longer).

I'll see if I can have a go at coding this – probably next week – and will share results and code...

PR #1407, opened 2019-02-05
Modernize debugging instructions

cwillisf:
This adds instructions and a few support files for building and running Scratch under Visual Studio Code, which is available for free. Previously, IntelliJ IDEA (not free) was the best-supported IDE for Scratch 2.0 development.

While we're generally not developing on Scratch 2.0 much these days, it's still nice to have as a reference when working on Scratch 3.0 features -- especially in terms of compatibility.

In preparation for adding the instructions I also cleaned up `README.md` a bit and corrected some out-of-date text.

I strongly recommend pressing the "Display the rich diff" button when reviewing `README.md`.

CC @mzgoddard @ktbee: these instructions now also work for the "online" version of Scratch 2.0

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,26 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "swf",
+            "request": "launch",
+            "name": "Launch SWF built by IDE",
+            "program": "${workspaceFolder}/build/ide/Scratch.swf"
+        },
+        {
+            "type": "swf",
+            "request": "launch",
+            "name": "Launch Flash 11.6 SWF built by Gradle",
+            "program": "${workspaceFolder}/build/11.6/Scratch.swf"
+        },
+        {
+            "type": "swf",
+            "request": "launch",
+            "name": "Launch Flash 10.2 SWF built by Gradle",
+            "program": "${workspaceFolder}/build/10.2/ScratchFor10.2.swf"
+        },
+    ]
+}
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
thisandagain, line 2: Are there any concerns here with this being invalid JSON?

thisandagain (approved):
Looks good! Just one concern listed inline.

README.md
```diff
@@ -1,30 +1,76 @@
...
+
+1. Build at least once using the Gradle instructions listed in the "Building" instructions below.
+   * TL;DR: run `./gradlew build -Ptarget="11.6"` in a terminal (on Windows, replace `/` with `\` as usual).
+   * This will download and unpack the necessary SDKs.
+   * You may need to agree to a few licenses (press `y` then `enter`).
+   * It may take quite a while and appear to hang at times. Watch disk and network activity to be sure.
+2. Install [Visual Studio Code](https://code.visualstudio.com/).
+3. Install the "ActionScript & MXML" extension (search for `@ext:as3` in the `Extensions` pane).
+   * Reload VS Code when prompted.
+4. Add the `scratch-flash-online` folder to the VS Code workspace.
```
towerofnix: I'm curious why this reads scratch-flash-online instead of scratch-flash -- I imagine it's info on how to merge the two repositories together, basically? Since scratch-flash-online is not a public repository and is not necessary to work on scratch-flash (e.g. in community modding), should this step be clarified as optional?

.vscode/launch.json
cwillisf: Ha, good question! This is a VS Code thing: they call it "JSONC" or "JSON with comments" but I guess they wanted to keep the `.json` file extension anyway. These comments are actually placed there by VS Code when you create the file.

See also: https://code.visualstudio.com/docs/languages/json#_json-with-comments

README.md
cwillisf: Good catch! I was doing these steps on both repositories at the same time and accidentally typed the wrong repository name here. I'll fix it :)

